PR #1, opened 2019-10-29
Improvements

Piccirello:
- Implement setup command
   - Consolidate print functions in print.go
   - Rename enable-env flag to no-read-env
   - Don't double-print the error when in debug mode
   - Swap table library to improve multi-line text support
   - Move argument parsing out of config
   - Add debug flag to print more info when encountering errors
  -  Don't re-parse value
 -   Consolidate logging logic
-    Make JSON a global flag
-  Move printing json to separate function

### DIFF
--- a/api/deploy.go
+++ b/api/deploy.go
@@ -58,7 +58,7 @@ func GetDeploySecrets(cmd *cobra.Command, apiKey string, project string, config 
 	params = append(params, utils.QueryParam{Key: "pipeline", Value: project})
 
 	host := cmd.Flag("deploy-host").Value.String()
-	response, err := utils.GetRequest(host, "v1/variables", params, apiKey)
+	response, err := utils.GetRequest(host, "/v1/variables", params, apiKey)
 	if err != nil {
 		utils.Log("Unable to fetch secrets")
 		return nil, err
@@ -76,7 +76,7 @@ func DownloadSecrets(cmd *cobra.Command, apiKey string, project string, config s
 	params = append(params, utils.QueryParam{Key: "metadata", Value: strconv.FormatBool(metadata)})
 
 	host := cmd.Flag("deploy-host").Value.String()
-	response, err := utils.GetRequest(host, "v1/variables", params, apiKey)
+	response, err := utils.GetRequest(host, "/v1/variables", params, apiKey)
 	if err != nil {
 		utils.Err(err, "Unable to download secrets")
 	}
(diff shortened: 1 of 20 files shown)


Piccirello:
This is a lot cleaner if you go by commit

api/deploy.go
```diff
@@ -61,7 +60,7 @@ func GetDeploySecrets(cmd *cobra.Command, apiKey string, project string, config
 	host := cmd.Flag("deploy-host").Value.String()
 	response, err := utils.GetRequest(host, "v1/variables", params, apiKey)
```
bvallelunga: Isn't it standard to have all api urls start with "/". So it would be "/v1/variables"

cmd/open.go
```diff
@@ -33,7 +33,7 @@ var openDashboardCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		err := open.Run("https://doppler.com")
 		if err != nil {
-			utils.Err(err)
+			utils.Err(err, "")
```
bvallelunga, line 36: What's with the empty string?

config/config.go
```diff
@@ -228,16 +210,16 @@ func Unset(scope string, options []string) {
 			scopedConfig := configContents[scope]
```
bvallelunga, line 210: seems like a lot of repeat code, can we abstract out?

utils/print.go
```diff
@@ -79,3 +82,171 @@ func PrintLog(log models.Log, jsonFlag bool) {
...
+}
+
+// PrintEnvironmentInfo print environment
+func PrintEnvironmentInfo(info models.EnvironmentInfo, jsonFlag bool) {
+	if jsonFlag {
+		PrintJSON(info)
+		return
+	}
+
+	rows := [][]string{{info.ID, info.Name, info.SetupAt, info.FirstDeployAt, info.CreatedAt, strings.Join(info.MissingVariables, ", "), info.Project}}
```
bvallelunga, line 142: can we make these multiline, looks hard to read on github

bvallelunga (approved):

cmd/open.go
Piccirello: Go doesn't support default values

utils/print.go
Piccirello: We can't, go-lint enforces the style

api/deploy.go
Piccirello: +1

config/config.go
Piccirello: I'll dedupe this code in the next pr

PR #9, opened 2019-11-13
Support bumping the version via release type: major/minor/patch

Piccirello:
Ex: `make release v=minor` will bump v1.4.2 to v1.5.0.

### DIFF
--- a/scripts/publish-deb.sh
+++ b/scripts/publish-deb.sh
@@ -21,6 +21,7 @@ if [ -z "$BINTRAY_API_KEY" ]; then
 fi
 
 getVersion () {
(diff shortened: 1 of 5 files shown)


scripts/publish-deb.sh
```diff
@@ -21,7 +21,7 @@ if [ -z "$BINTRAY_API_KEY" ]; then
 fi
 
 getVersion () {
-  VERSION=$(git describe);
+  VERSION=$(git describe --abbrev=0);
```
bvallelunga, line 25: Can we add a note to explain what the `--abbrev=0` does and why it matters?

bvallelunga (approved):

PR #10, opened 2019-11-15
Support short flags for key, project, and config

Piccirello:


pkg/cmd/environments.go
```diff
@@ -57,7 +57,7 @@ var environmentsGetCmd = &cobra.Command{
 }
 
 func init() {
-	environmentsGetCmd.Flags().String("project", "", "output json")
+	environmentsGetCmd.Flags().StringP("project", "p", "", "output json")
```
bvallelunga: shouldn't the description be "doppler project"?

pkg/cmd/configs.go
```diff
@@ -192,46 +192,46 @@ var configsLogsRollbackCmd = &cobra.Command{
 }
 
 func init() {
-	configsCmd.Flags().String("project", "", "doppler project (e.g. backend)")
+	configsCmd.Flags().StringP("project", "p", "", "doppler project (e.g. backend)")
```
bvallelunga: can we change all the "doppler project" to "enclave project"?

pkg/cmd/environments.go
Piccirello: good catch

pkg/cmd/root.go
```diff
@@ -60,7 +60,7 @@ func init() {
 	rootCmd.Version = version.ProgramVersion
 	rootCmd.SetVersionTemplate(rootCmd.Version + "\n")
 
-	rootCmd.PersistentFlags().String("key", "", "doppler api key")
+	rootCmd.PersistentFlags().StringP("key", "k", "", "doppler api key")
```
bvallelunga: if we are moving to using the word token instead of key, should we switch the "k" to "t"?

Piccirello: Sure, I'll switch over to using `token` throughout. We'll have to manually update our existing config files. Alternatively, you can delete your config and let the migration script run again.

bvallelunga: Dope I will do that.

pkg/cmd/root.go
```diff
@@ -60,12 +60,12 @@ func init() {
 	rootCmd.Version = version.ProgramVersion
 	rootCmd.SetVersionTemplate(rootCmd.Version + "\n")
 
-	rootCmd.PersistentFlags().String("key", "", "doppler api key")
+	rootCmd.PersistentFlags().StringP("token", "t", "", "doppler api key")
```
bvallelunga: Change "doppler api key" with "doppler token"

bvallelunga (approved):

PR #18, opened 2019-12-05
Set dashboard url from server during login flow

Piccirello:
- Set dashboard url from server during login flow
- Sort cli configuration before printing
- Replace duplication with iteration

### DIFF
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -133,6 +133,18 @@ func LocalConfig(cmd *cobra.Command) models.ScopedConfig {
 		}
 	}
 
+	flagSet = cmd.Flags().Changed("dashboard-host")
+	if flagSet || localConfig.DashboardHost.Value == "" {
+		localConfig.DashboardHost.Value = cmd.Flag("dashboard-host").Value.String()
+		localConfig.DashboardHost.Scope = "*"
+
+		if flagSet {
+			localConfig.DashboardHost.Source = models.FlagSource.String()
+		} else {
+			localConfig.DashboardHost.Source = models.DefaultValueSource.String()
+		}
+	}
+
 	flagSet = cmd.Flags().Changed("no-verify-tls")
 	if flagSet || localConfig.VerifyTLS.Value == "" {
 		noVerifyTLS := cmd.Flag("no-verify-tls").Value.String()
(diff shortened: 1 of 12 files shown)


pkg/configuration/config.go
```diff
@@ -324,25 +290,17 @@ func parseScope(scope string) (string, error) {
 
 // IsValidConfigOption whether the specified key is a valid option
 func IsValidConfigOption(key string) bool {
-	return key == "token" || key == "project" || key == "config" || key == "api-host" || key == "verify-tls"
+	return key == "token" || key == "project" || key == "config" || key == "api-host" ||
```
bvallelunga: Can we put this those strings in a set and check if the key exists in the set? I think that would be much cleaner and easier to extend in the future.

pkg/utils/print.go
```diff
@@ -371,23 +343,26 @@ func PrintConfigs(configs map[string]models.Config, jsonFlag bool) {
...
+	// sort by scope, then by name
+	sort.Slice(rows, func(a, b int) bool {
+		if rows[a][2] > rows[b][2] {
+			return false
 		}
-		if config.VerifyTLS != "" {
-			rows = append(rows, []string{"verify-tls", config.VerifyTLS, scope})
+		if rows[a][2] < rows[b][2] {
+			return true
 		}
```
bvallelunga: I think we can clean up this code with

``` go
if rows[a][2] != rows[b][2] {
    return rows[a][2] < rows[b][2] 
}
```

bvallelunga (approved):

pkg/utils/print.go
Piccirello: +1

pkg/configuration/config.go
Piccirello: This is the equivalent code in Go. I don't think it's worth it:

```golang
	configOptions := [...]string{
		"token",
		"project",
		"config",
		"api-host",
		"dashboard-host",
		"verify-tls",
	}

	for _, option := range configOptions {
		if key == option {
			return true
		}
	}

	return false
```

bvallelunga: What about:

``` go
configOptions := [...]string{
		"token",
		"project",
		"config",
		"api-host",
		"dashboard-host",
		"verify-tls",
}

val, exists := dict["foo"]
return exists
```

Reference
https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go

Piccirello: +1

PR #28, opened 2019-12-08
Timeout incomplete login flow after 5 minutes

Piccirello:
This is in addition to our server-side timeouts

### DIFF
--- a/pkg/cmd/login.go
+++ b/pkg/cmd/login.go
@@ -68,8 +68,17 @@ var loginCmd = &cobra.Command{
 			fmt.Println("Waiting...")
 		}
 
+		// auth flow must complete within 5 minutes
+		timeout := 5 * time.Minute
+		completeBy := time.Now().Add(timeout)
+
 		response = nil
 		for {
+			// we don't respect --no-timeout here
+			if time.Now().After(completeBy) {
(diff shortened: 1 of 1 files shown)


pkg/cmd/login.go
```diff
@@ -68,8 +68,17 @@ var loginCmd = &cobra.Command{
...
 
+		// auth flow must complete within 5 minutes
+		timeout := 5 * time.Minute
+		completeBy := time.Now().Add(timeout)
+
 		response = nil
 		for {
+			// we don't respect --no-timeout here
+			if time.Now().After(completeBy) {
+				utils.HandleError(fmt.Errorf("login timed out after %d minutes", int(timeout.Minutes())))
```
bvallelunga, line 79: Can we capitalize "login"?

bvallelunga (approved):

pkg/cmd/login.go
Piccirello: This is lowercase for consistency with our other errors. `HandleError()` will prepend `Error:` to this, so the printed text will read `Error: login timed out after 5 minutes`.

bvallelunga: gotcha

PR #47, opened 2019-12-12
Don't print version update message if plain flag is specified 

Piccirello:
- Don't check for CLI updates if we can't print the results


Piccirello:
This was causing our `radar-server` build to break because `doppler enclave secrets get NPM_TOKEN --plain` was printing this:

```sh
Doppler CLI version 1.0.0 is now available
THE-NPM-TOKEN
```

instead of just 

```sh
THE-NPM-TOKEN
```

bvallelunga (approved):

PR #54, opened 2020-01-06
Print service token on creation

Piccirello:
When creating a service token we now print an additional column: `Token`

### DIFF
--- a/pkg/printer/print.go
+++ b/pkg/printer/print.go
@@ -510,8 +510,8 @@ func ConfigOptionNames(options []string, jsonFlag bool) {
 	Table([]string{"name"}, rows, TableOptions())
 }
 
-// ConfigServiceTokens print config service tokens
-func ConfigServiceTokens(tokens []models.ConfigServiceToken, number int, jsonFlag bool) {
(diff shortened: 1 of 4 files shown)


pkg/printer/print.go
```diff
@@ -527,12 +527,23 @@ func ConfigServiceTokens(tokens []models.ConfigServiceToken, number int, jsonFla
 	Table([]string{"name", "slug", "created at"}, rows, TableOptions())
 }
 
-// ConfigServiceToken print config service token
+// ConfigServiceTokenInfo print config service token info
+func ConfigServiceTokenInfo(token models.ConfigServiceToken, jsonFlag bool, diff bool) {
+	if jsonFlag {
```
bvallelunga, line 532: Is there a way we can add a --plain option or some flag just to return the token?

bvallelunga (changes requested):

pkg/printer/print.go
Piccirello: I like it!

bvallelunga (approved):

PR #56, opened 2020-01-13
Provide more options for downloading secrets

Piccirello:
Secrets can now be downloaded in json format, in addition to env format. The downloaded secrets can also now printed to stdout instead of a file.

Usage:
```
$ doppler enclave secrets download && cat doppler.env
# This is the metadata
A="B"

$ doppler enclave secrets download --json && cat doppler.json
{
  "A": "B"
}

$ doppler enclave secrets download --no-file
# This is the metadata
A="B"

$ doppler enclave secrets download --json --no-file
{
  "A": "B"
}
```

### DIFF
--- a/pkg/cmd/enclave_secrets.go
+++ b/pkg/cmd/enclave_secrets.go
@@ -16,6 +16,7 @@ limitations under the License.
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


bvallelunga:
I thought we were going to remove the download command?

bvallelunga:
can we add yaml support?

Piccirello:
> I thought we were going to remove the download command?

I thought about it some more and decided against it. Let's discuss tomorrow- maybe I'm wrong.

> can we add yaml support?

I like this idea, but before we implement do we have a use case for this?

bvallelunga:
Also if we call the download command without a path, will it use `doppler-backup` as the file name?

Piccirello:
> will it use doppler-backup as the file name?

Updated the PR description- it'll be called `doppler` as before. I'd really prefer `doppler-backup`, or possible even something like `HASH(PROJECT+CONFIG)-backup.env`, but that's a breaking change and will require a major bump.

pkg/cmd/enclave_secrets.go
```diff
@@ -95,11 +94,10 @@ var secretsSetCmd = &cobra.Command{
 	Long: `Set the value of one or more secrets.
 
 Ex: set the secrets "api_key" and "crypto_key":
-doppler secrets set api_key=123 crypto_key=456`,
+doppler enclave secrets set api_key=123 crypto_key=456`,
```
bvallelunga: can we use example secrets where the name is all caps?

bvallelunga (approved):

pkg/cmd/enclave_secrets.go
Piccirello: updated the examples

Piccirello:
Re-requesting due to using a new api endpoint and removing some code

pkg/http/api.go
```diff
@@ -135,14 +135,19 @@ func RevokeAuthToken(host string, verifyTLS bool, token string) (map[string]inte
...
 // DownloadSecrets for specified project and config
-func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string, config string, metadata bool) ([]byte, Error) {
+func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string, config string, metadata bool, jsonFlag bool) ([]byte, Error) {
 	var params []queryParam
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "metadata", Value: strconv.FormatBool(metadata)})
 
 	headers := apiKeyHeader(apiKey)
-	headers["Accept"] = "text/plain"
-	statusCode, response, err := GetRequest(host, verifyTLS, headers, "/enclave/v1/configs/"+config+"/secrets", params)
```
Piccirello, line 145: Shoot, this is actually a breaking change because the new endpoint ignores the `metadata` query param (which also means I should remove all `metadata` code). How do you want to handle this- make all our discussed changes "breaking", or temporarily support both endpoints so that v1 can get json download support?

If no one is actually asking for the json feature, I say we just do a big v2 release with all these changes. The affected user base has to be extremely small still.

bvallelunga: I say lets do a big v2 release. I think that is the safest bet.

pkg/cmd/enclave_secrets.go
```diff
@@ -164,26 +161,48 @@ doppler secrets delete api_key crypto_key`,
 var secretsDownloadCmd = &cobra.Command{
 	Use:   "download <filepath>",
 	Short: "Download a config's .env file",
-	Long: `Save your config's secrets to a .env file. The default filepath is ./doppler.env
+	Long: `Download your config's secrets for later use. Env and JSON format are supported.
+
+Examples:
+
+Save your secrets to /root/ in Env format
+$ doppler enclave secrets download /root/secrets.env
```
bvallelunga, line 170: Can we make json the default? I really think we should move Doppler and it's customers away from this legacy format. I recommend having a `--format=env` flag. This would enable us to support future formats.

bvallelunga (changes requested):
Marking as "request changes" to acknowledge feedback.

pkg/cmd/enclave_secrets.go
```diff
@@ -210,22 +228,21 @@ func init() {
...
 	secretsDeleteCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
 	secretsDeleteCmd.Flags().Bool("silent", false, "do not output the response")
 	secretsDeleteCmd.Flags().Bool("yes", false, "proceed without confirmation")
 	secretsCmd.AddCommand(secretsDeleteCmd)
 
 	secretsDownloadCmd.Flags().StringP("project", "p", "", "enclave project (e.g. backend)")
 	secretsDownloadCmd.Flags().StringP("config", "c", "", "enclave config (e.g. dev)")
-	secretsDownloadCmd.Flags().Bool("no-metadata", false, "do not add metadata to the downloaded file (helps cache busting)")
+	secretsDownloadCmd.Flags().Bool("no-file", false, "print the response to stdout; don't save to a file")
+	secretsDownloadCmd.Flags().Bool("no-metadata", false, "do not add metadata to the downloaded file (metadata can be used for cache busting)")
```
bvallelunga: Should we remove this flag?

Piccirello: Addressed above

pkg/cmd/enclave_secrets.go
Piccirello: Will switch json to the default when we add encryption, because then it won't really matter what the format is.

bvallelunga (approved):

PR #59, opened 2020-01-16
Create salus.yaml

Piccirello:


bvallelunga (approved):

bvallelunga (commented):
Please add a name to the job

.github/workflows/salus.yaml
```diff
@@ -0,0 +1,19 @@
+on: [push]
+
+jobs:
+  salus_scan_job:
+    runs-on: ubuntu-latest
+    name: Salus Security Scan
+    steps:
+    - uses: actions/checkout@v1
```
ScottBrenner:
```suggestion
    - uses: actions/checkout@v2
```
`v2` is the latest https://github.com/actions/checkout#whats-new

Piccirello: Thanks for the suggestion!

PR #60, opened 2020-01-17
Enable the run command's fallback file by default

Piccirello:
By default, the fallback file will be written to the user's configuration directory. Fallback files are scoped per (project, config). Files downloaded with the `enclave secrets download` command can also now be used as fallback files.

We don't yet have encryption, but that'll be enabled before we release this.

When the CLI hits an error with writing the fallback file, this is what the user sees:

Screen Shot 2020-01-16 at 10 31 37 PM

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
 	"github.com/DopplerHQ/cli/pkg/http"
@@ -34,6 +35,8 @@ import (
 
 // DefaultFallbackDir path to the default fallback dir
 var DefaultFallbackDir string
+
+var defaultFallbackFileMaxAge = 72 * time.Hour
 
 var runCmd = &cobra.Command{
 	Use:   "run [command]",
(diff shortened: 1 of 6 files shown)


pkg/cmd/run.go
```diff
@@ -74,67 +102,111 @@ doppler run --token=123 -- printenv`,
...
+func writeFailureMessage() []string {
+	var msg []string
+
+	msg = append(msg, "")
+	msg = append(msg, color.Green.Render("Why did doppler exit?"))
+	msg = append(msg, "Doppler failed to make a local backup of your secrets, known as a fallback file.")
+	msg = append(msg, "The most common cause for this is insufficient permissions, including using a fallback file created by a different user.")
+	msg = append(msg, "")
+	msg = append(msg, color.Green.Render("Why does that matter?"))
+	msg = append(msg, "Without the fallback file, your secrets would be inaccessible in the event of a network partition or Doppler downtime.")
```
bvallelunga: `Without the fallback file, your secrets would be inaccessible in the event you do not have internet access or Doppler downtime.`

bvallelunga (approved):

pkg/cmd/run.go
Piccirello: I much prefer the current messaging, especially in a developer tool

Piccirello: Renamed `network partition` to `network outage`

PR #61, opened 2020-01-17
Organize all user config within a 'doppler' directory

Piccirello:
This is to help organize our files as we put more things in the config directory, like fallback files.
This change includes automatic config migration for CLI v1 users, as well as a continued migration path for node cli users.

Example:
before /Users/thomas/Library/Application Support/.doppler.yaml
after /Users/thomas/Library/Application Support/doppler/.doppler.yaml

**I also plan to add support for this new location in CLI v1 to maintain backwards compatibility.**


bvallelunga (approved):

Piccirello:
Before merging: change branch

PR #63, opened 2020-01-18
Encrypt all files containing secrets

Piccirello:
Fallback files and downloaded secrets will now be encrypted before touching the filesystem.
PBKDF2 is used against the (token, project, config) to generate the key.
AES-GCM is used with a random salt to encrypt the file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/pflag v1.0.5 // indirect
 	go.mongodb.org/mongo-driver v1.1.2 // indirect
+	golang.org/x/crypto v0.0.0-20190530122614-20be4c3c3ed5
 	golang.org/x/sys v0.0.0-20191029155521-f43be2a4598c // indirect
 	gopkg.in/gookit/color.v1 v1.1.6
 	gopkg.in/yaml.v3 v3.0.0-20191026110619-0b21df46bc1d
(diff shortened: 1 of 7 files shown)


Piccirello:
In this version there's no way to disable encryption. You can provide a custom encryption passphrase, however.

If we determine that an option for disabling encryption is a must, I can add it in. Let's discuss tomorrow.

bvallelunga:
Got a couple questions since it's hard to follow along in the code:

1. Does this also encrypt files downloaded with the `doppler enclave secrets download` command?
      - If yes, is there a way to download it unencrypted? From the comment above I'm guessing new fallback files created by the run command will always be encrypted. 
2. If a plain text fallback file is provided, will the CLI handle it?

Piccirello:
> Does this also encrypt files downloaded with the `doppler enclave secrets download` command? 

Yes

> is there a way to download it unencrypted?

Not currently. However users can still use the `--no-file` flag to print the unencrypted secrets to stdout, which is meant to be piped into another running application.

> If a plain text fallback file is provided, will the CLI handle it?

Not in the current iteration. Plaintext support is completely removed. There's probably a case for supporting it, however I'm not sure how users would get an unencrypted copy of their secrets in our non-standard json format.


Our goal is to prevent unencrypted secrets from ever touching the filesystem, and I think we should try **really** hard to accomplish that goal. I think the best way to accomplish it is not adding that option unless our customers present us with a really strong use case.

bvallelunga (approved):

PR #65, opened 2020-01-20
Fix bug with revoking previous token after login

Piccirello:


pkg/cmd/login.go
```diff
@@ -204,9 +204,12 @@ This is an alias of the "logout" command.`,
...
 		updateConfig := !utils.GetBoolFlag(cmd, "no-update-config")
 		verifyTLS := utils.GetBool(localConfig.VerifyTLS.Value, true)
+		yes := utils.GetBoolFlag(cmd, "yes")
 		token := localConfig.Token.Value
 
-		revokeToken(localConfig.APIHost.Value, token, silent, verifyTLS, updateConfig)
+		if yes || utils.ConfirmationPrompt(fmt.Sprintf("Revoke auth token scoped to %s?", localConfig.Token.Scope), false) {
+			revokeToken(localConfig.APIHost.Value, token, silent, verifyTLS, updateConfig)
+		}
 	},
```
bvallelunga: Can we check if a previous token is set for that scope before prompting the user?

Also can we rephrase to state this is the previous auth token. Something like "Looks like you already have an auth token for this scope, do you want to revoke it before creating the new one?"

bvallelunga (approved):

pkg/cmd/login.go
Piccirello: To clarify, this prompt is for the `login revoke` command.

PR #66, opened 2020-01-24
Support copying values to the clipboard

Piccirello:
Currently supports copying from:
- enclave secrets get
- configs tokens create
- configure get

These are all places we have the `--plain` flag.


bvallelunga:
What was the inspiration for this PR? Could you share what the use case is?

Piccirello:
> What was the inspiration for this PR? Could you share what the use case is?

Sure thing. This is a feature I've been wanting when using `doppler enclave secrets get SECRET_NAME --plain` or `doppler configure get token --plain`. When running those commands I'm almost exclusively doing so so that I can copy the value. With a copy flag, I can perform this action directly without having to break my flow by using the mouse/trackpad.

bvallelunga (approved):

PR #70, opened 2020-01-29
Support parsing home directory-based paths

Piccirello:
In some cases, like `--scope=PATH`, the shell won't parse the path. This means we'll need to manually parse the path ourselves, like expanding `~/test` to `/Users/thomas/test`.

For example, calling `doppler enclave secrets download ~/doppler.json` works, but `doppler enclave secrets download "~/doppler.json"` results in an error `open ~/doppler.json: no such file or directory`.

The one case we don't (and can't easily) support parsing is a path like `~USERNAME/` where `USERNAME` is NOT the current user. In that case we detect our inability to parse it and print an error asking the user to provide an absolute path.

### DIFF
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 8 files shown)


pkg/utils/util.go
```diff
@@ -55,6 +56,34 @@ func HomeDir() string {
...
+	}
+
+	absolutePath, err := filepath.Abs(filepath.Clean(path))
+	if err != nil {
+		return "", err
+	}
+
+	return absolutePath, nil
+}
+
```
bvallelunga: Maybe a cleaner way to write this:

``` js
function parsePath(input) {
  if(input.hasPrefix("~")) {
    const first_path = input.split("/")[0]

    if(first_path !== "~" && first_path !== `~${USERNAME}`) {
      throw new Error("We don't support relative paths to point to other users!")
    }

    input.replace(first_path, ABSOLUTE_HOME_DIR_PATH)
  }

  return input
}
```

pkg/utils/util_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Error(fmt.Sprintf("Got %s, expected %s", path, homeDir))
+	}
+
+	// expect error
+	path, err = ParsePath(fmt.Sprintf("~%s/", username+"1"))
+	if err == nil || path != "" {
+		t.Error(fmt.Sprintf("Got %s, expected error", path))
+	}
+}
+
```
bvallelunga, line 80: More tests:

- `~/~`
- `~user/~otheruser`
- `~otheruser/~user`
- `~user/content/to/path/~`

bvallelunga (approved):

pkg/utils/util_test.go
Piccirello: This function is meant as a helper to parse a leading tilde, rather than as a strict validator against invalid syntax.

pkg/utils/util.go
Piccirello: Re-implemented using your implementation

bvallelunga (approved):

PR #76, opened 2020-03-27
Fix enclave setup not overwriting the previous config

Piccirello:


pkg/cmd/enclave_setup.go
```diff
@@ -116,17 +116,23 @@ var setupCmd = &cobra.Command{
...
+			var options []string
+			var defaultOption string
 			for _, val := range configs {
-				configOptions = append(configOptions, val.Name)
+				option := val.Name
+				options = append(options, option)
+				// reselect previously-configured value
+				if val.Name == scopedConfig.EnclaveConfig.Value {
+					defaultOption = val.Name
+				}
```
bvallelunga: Why should we care about the previous value?

Piccirello: So that we can re-select the previous value when re-running setup. I believe this was a feature you requested.

bvallelunga: Gotcha, can we only re-select the config if the project is the same.

Piccirello: I like that

Piccirello: Addressed

bvallelunga (approved):

PR #82, opened 2020-05-06
Allow for brew upgrades via "doppler update"

Piccirello:


pkg/cmd/update.go
```diff
@@ -0,0 +1,80 @@
...
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "update the Doppler CLI",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		if installedViaBrew() {
+			brewUpdate()
+		} else {
+			utils.HandleError(fmt.Errorf("this command is not yet implemented for your install method"))
+		}
```
bvallelunga, line 36: I think we can make this more modular for future package managers. Since I don't really know Go im going to write the suggestion in javascript.

``` js
switch(true) {
    case installedViaBrew(): return brewUpdate()
    case installedViaAPK(): return apkUpdate()
    default: throw new Error("this command is not yet implemented for your install method")
}
```

bvallelunga (dismissed):

pkg/cmd/update.go
Piccirello: My plan was to build it out as we support more. I'm just keeping it dead simple for now.

bvallelunga: okay sounds good.

bvallelunga (approved):

PR #83, opened 2020-05-07
Increase the verbosity of the doppler run command

Piccirello:
More errors are printed by default, and existing printed errors contain more information.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -181,6 +181,11 @@ var runCleanCmd = &cobra.Command{
 		deleted := 0
 		now := time.Now()
 
+		action := "Deleted"
+		if dryRun {
+			action = "Would have deleted"
+		}
+
 		for _, entry := range entries {
 			if entry.IsDir() {
 				continue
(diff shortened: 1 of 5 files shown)


Piccirello:
I'd recommend upgrading to the latest CLI version before this is released. Then you'll be able to pick up these changes with `doppler update` 😃

pkg/cmd/run.go
```diff
@@ -360,7 +361,8 @@ func init() {
 	runCmd.Flags().Bool("fallback-readonly", false, "disable modifying the fallback file. secrets can still be read from the file.")
 	runCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
 	runCmd.Flags().Bool("no-exit-on-write-failure", false, "do not exit if unable to write the fallback file")
-	runCmd.Flags().Bool("silent", false, "disable error output if the supplied command exits non-zero")
+	runCmd.Flags().Bool("silent", false, "disable output of info messages")
+	runCmd.Flags().Bool("silent-exit", false, "disable error output if the supplied command exits non-zero")
```
bvallelunga, line 369: I like this!

pkg/http/github.go
```diff
@@ -52,12 +52,8 @@ func CheckCLIVersion(versionCheck models.VersionCheck, silent bool, json bool, d
...
 	tag, err := getLatestVersion()
 	if err != nil {
-		if debug && !json {
-			utils.LogError(err)
-		}
-		if !silent && !json {
-			utils.LogError(errors.New("Unable to check for CLI updates"))
-		}
+		utils.LogError(errors.New("Unable to check for CLI updates"))
+		utils.LogDebugError(err)
```
bvallelunga, line 56: I think is this a breaking change. We will have to notify all our customers to now use the `--silent` flag.

bvallelunga (approved):

pkg/http/github.go
Piccirello: This isn't a breaking change. This command already uses the `--silent` flag.

pkg/cmd/run.go
Piccirello: This is actually a breaking change. But since this was just released two days ago and only one customer is using it, I think it's alright. I'll notify them directly.

pkg/http/github.go
bvallelunga: Sorry what I mean is that users who are using the `--json` flag are expecting no logging on CLI update failures. Now they will need to provide `--silent` flag to get the same effect.

Piccirello: Ah gotcha. It's true that this now prints when running with `--json` and `-debug`, however it was already standard practice to always print debug messages, even when using `--json`. Thus this was the only place that didn't follow this. Therefore the breakiness is a little fuzzy to me because anyone calling `--json --debug` could already expect to see non-json debug messages.

bvallelunga: Okay that makes sense. I think it is fine then.

PR #84, opened 2020-05-08
Auto-select sole project/config during 'enclave setup'

Piccirello:
- Auto-select sole project/config during 'enclave setup'
- Save token during 'enclave setup` when passed via flag/env

With these two changes, it is now sufficient to run `doppler enclave setup --token=$TOKEN --no-prompt` to fully configure a service token. The project and config are no longer required, and neither is the call to `doppler configure`.

### DIFF
--- a/pkg/cmd/enclave_setup.go
+++ b/pkg/cmd/enclave_setup.go
@@ -37,12 +37,24 @@ var setupCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		silent := utils.GetBoolFlag(cmd, "silent")
 		promptUser := !utils.GetBoolFlag(cmd, "no-prompt")
+		canSaveToken := !utils.GetBoolFlag(cmd, "no-save-token")
 		scope := cmd.Flag("scope").Value.String()
 		localConfig := configuration.LocalConfig(cmd)
 		scopedConfig := configuration.Get(scope)
 
 		utils.RequireValue("token", localConfig.Token.Value)
 
+		saveToken := false
+		if canSaveToken {
+			// save the token when it's passed via command line
+			switch localConfig.Token.Source {
+			case models.FlagSource.String():
+				saveToken = true
+			case models.EnvironmentSource.String():
+				utils.Log(valueFromEnvironmentNotice("DOPPLER_TOKEN"))
+				saveToken = true
+			}
+		}
 
 		currentProject := localConfig.EnclaveProject.Value
 		selectedProject := ""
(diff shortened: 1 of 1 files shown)


pkg/cmd/enclave_setup.go
```diff
@@ -118,6 +118,14 @@ func selectProject(projects []models.ProjectInfo, prevConfiguredProject string,
 		}
 	}
 
+	if len(projects) == 1 {
+		// the user is expecting to a prompt, so print a message instead
+		if promptUser {
+			utils.Log("Selected only available project")
```
bvallelunga: We should show the project name:

```
Selected only available project: Monolith
```

We could also bold the `Selected only available project:` part of the message.

I find it a little odd that I didn't see or select which project was selected. You can replicate this by having a workplace with a single project. I am using the `Doppler Demo` workpalce.

Screen Shot 2020-05-08 at 7 23 16 PM

pkg/cmd/enclave_setup.go
```diff
@@ -158,6 +166,14 @@ func selectConfig(configs []models.ConfigInfo, selectedConfiguredProject bool, p
 		}
 	}
 
+	if len(configs) == 1 {
+		// the user is expecting to a prompt, so print a message instead
+		if promptUser {
+			utils.Log("Selected only available config")
```
bvallelunga: We should show the config name:

```
Selected only available config: dev
```

We could also bold the `Selected only available config:` part of the message.

bvallelunga (changes requested):

pkg/cmd/enclave_setup.go
Piccirello: > We should show the project name:

The reason we don't is because it's shown when we print the config.

Piccirello: > We could also bold the Selected only available project: part of the message.

I think we should treat this like any other log message

> I find it a little odd that I didn't see or select which project was selected.

See the log message about it selecting the only available project

pkg/cmd/enclave_setup.go
Piccirello: See my other comment

pkg/cmd/enclave_setup.go
bvallelunga: Isn't the project id being shown, not the name of the project. I honestly found it a little jarring when I ran it and didn't know which project was auto-selected.

Piccirello: I think this is the one use case where it makes sense to show it. I'll add the name in the message.

Piccirello: Changes pushed, try out the latest version

bvallelunga: Testing!

bvallelunga: Nit: Can we change the color to a notice/warning color like orange and then bold the `Selected only available project:` part?

Piccirello: The reason I'm avoiding putting emphasis on the lines is that I don't think we should need emphasis. If it's printed, it's important and the user should read it. If it's not important enough to be bolded, it shouldn't be printed at all.

bvallelunga: To be clear, I am not saying bold the entire line. Just the `Selected only available project:` part. 

For example the line could look like this:
**Selected only available project:** Partners (ep_d14e2b0a7b1)

I think this helps make it easier on the eye to see the message separated from the selected project. I know it's a nit but I think it would help.

Piccirello: I follow what you're saying, I'm just not sure I agree. I pushed up a change with the bolded text so you can try it out.

bvallelunga: I like it! It is not as significant as I thought it would be but I do like it.

bvallelunga (approved):

PR #88, opened 2020-05-10
Scope login/logout to the current directory

Piccirello:
The previous default was global.

### DIFF
--- a/pkg/cmd/login.go
+++ b/pkg/cmd/login.go
@@ -43,6 +43,10 @@ var loginCmd = &cobra.Command{
 		copyAuthCode := !utils.GetBoolFlag(cmd, "no-copy")
 		hostname, _ := os.Hostname()
 
+		if !cmd.Flags().Changed("scope") {
+			utils.LogWarning("Your token will be scoped to the current directory and its subdirectories. You may specify an alternative directory with the --scope flag.")
+		}
+
 		response, err := http.GenerateAuthCode(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), hostname, utils.HostOS(), utils.HostArch())
 		if !err.IsNil() {
 			utils.HandleError(err.Unwrap(), err.Message)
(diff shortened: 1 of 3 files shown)


bvallelunga:
Why are we making this change? A while back we agreed on keeping the token globally scoped.

Piccirello:
I've always really hated our default scoping, to the point where all of our login examples include specifying a manual `--scope`. I only agreed to using a default because we didn't have a good way of communicating the scope to the user. But now we simply log when they're using a default scope.

Anyone using `doppler enclave setup ` is already scoping their project/config to the current directory, and if passing it a token then the token is scoped to the current directory too.

bvallelunga:
> I've always really hated our default scoping, to the point where all of our login examples include specifying a manual `--scope`. I only agreed to using a default because we didn't have a good way of communicating the scope to the user. But now we simply log when they're using a default scope.
> 
> Anyone using `doppler enclave setup ` is already scoping their project/config to the current directory, and if passing it a token then the token is scoped to the current directory too.

So I think we are mixing up a couple behaviours. I see there are 3:

- Us doing demos (where we always want the scope)
- Doppler CLI used in docker or other production settings (scope doesn't matter)
- Local development performed by our users (token should be globally scoped)

That being said, I think local development is our primary usecase for the login flow. Since all users are part of one workplace right now (except us), it makes sense that the token would be globally scoped. I think scoping adds mental overhead which is why it's important that the default settings be set in a way that reduces any mental overhead. The login command being globally scoped makes as a default, while the enclave setup command makes sense for it to be scoped to the users current directory. In both example, I think the default scope is correct for the usecase.

Piccirello:
> Since all users are part of one workplace right now (except us), it makes sense that the token would be globally scoped. 

I don't agree with you at all on this topic! But this line has convinced me. Until we have users in multiple workplaces (and thus multiple tokens), it doesn't benefit the user much to narrow the default scope. I do think this should be revisited when users are in multiple workplaces, or if we ever change our auth model to scope CLI tokens to fewer resources.

PR #90, opened 2020-05-14
 Add CLI install script 

Piccirello:
This script automatically fetches and installs the latest Doppler CLI. It can be downloaded from https://cli.doppler.com/install.sh

Documentation:
- [x] docs.doppler.com
- [x] readme.md
- [x] install.md
- [ ] notify customers (incl. changelog post)

After merge:
- [x] Update Cloudflare worker install.sh url to point to master


Piccirello:
This install script needs to be documented:

- [ ] docs.doppler.com
- [ ] readme.md
- [ ] install.md
- [ ] notify customers (incl. changelog post)

scripts/install.sh
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env sh
+
+set -e
```
bvallelunga, line 3: Why is `set -e` needed?

bvallelunga (approved):

scripts/install.sh
Piccirello: `set -e` tells the shell to exit if any command returns non-zero. Without it, this script would continue executing even if the download failed:

```sh
curl -O https://cli.download.com/install.sh
ls install.sh
```

bvallelunga: Makes sense, that is pretty cool!

PR #91, opened 2020-05-15
Print invalid JSON responses in debug mode

Piccirello:
And other small changes.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -39,14 +39,19 @@ var rootCmd = &cobra.Command{
 		configuration.Setup()
 		configuration.LoadConfig()
 
+		silent := utils.GetBoolFlagIfChanged(cmd, "silent", false)
+		plain := utils.GetBoolFlagIfChanged(cmd, "plain", false)
+
 		if utils.Debug {
+			if silent {
+				utils.LogWarning("--silent has no effect when used with --debug")
+			}
+
 			utils.LogDebug("Active configuration")
 			printer.ScopedConfigSource(configuration.LocalConfig(cmd), false, true)
 			fmt.Println("")
 		}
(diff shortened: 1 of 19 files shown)


pkg/cmd/root.go
```diff
@@ -40,7 +40,6 @@ var rootCmd = &cobra.Command{
 		configuration.LoadConfig()
```
bvallelunga, line 40: I am guessing now users will need to provide the `--silent` flag too so their flows don't break?

Piccirello: Sorry, I don't follow

bvallelunga: Based on this commit. We now print debug/warning/cli upgrade messages when using `--plain` which will break a users our flow. For example we fetch the ngrok domain using `--plain`. If a upgrade Doppler warning was printed that would break our ngrok flow. To fix this I am guessing we need to use the `--silent` flag.

Piccirello: Gotcha, FYI GitHub doesn't show me what commit you commented on. I didn't fully consider the implication of this change, but I suppose it is a breaking change. I don't think this tiny change is worth another major version bump so I'll remove this. I still think this is a good idea long term to help standardize flag behavior.

bvallelunga: I'm gonna approve this PR assuming this commit gets dropped. Commit: 351dffba6c4fb8c2465f3f00687f439fc8cfd222

bvallelunga (approved):

pkg/cmd/root.go
Piccirello: Changes pushed with a different commit

bvallelunga (approved):

PR #99, opened 2020-06-01
Print active configuration when specifying --print-config flag

Piccirello:
- Don't force print stacktraces in debug mode 
- Simplify logic to catch panics in non-dev builds 
- Use application defaults rather than hardcoded defaults

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -44,8 +44,12 @@ var rootCmd = &cobra.Command{
 			if silent {
 				utils.LogWarning("--silent has no effect when used with --debug")
 			}
+		}
 
-			utils.LogDebug("Active configuration")
+		// this output does not honor --silent
+		printConfig := utils.GetBoolFlagIfChanged(cmd, "print-config", false)
+		if printConfig {
+			fmt.Println("Active configuration")
 			printer.ScopedConfigSource(configuration.LocalConfig(cmd), false, true)
 			fmt.Println("")
 		}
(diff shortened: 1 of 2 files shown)


pkg/cmd/root.go
```diff
@@ -138,7 +141,8 @@ func init() {
 	rootCmd.PersistentFlags().Bool("no-read-env", false, "do not read enclave config from the environment")
 	rootCmd.PersistentFlags().String("scope", ".", "the directory to scope your config to")
 	rootCmd.PersistentFlags().String("configuration", configuration.UserConfigFile, "config file")
-	rootCmd.PersistentFlags().Bool("json", false, "output json")
-	rootCmd.PersistentFlags().Bool("debug", false, "output additional information when encountering errors")
-	rootCmd.PersistentFlags().Bool("silent", false, "disable output of info messages")
+	rootCmd.PersistentFlags().Bool("json", utils.OutputJSON, "output json")
+	rootCmd.PersistentFlags().Bool("debug", utils.Debug, "output additional information")
+	rootCmd.PersistentFlags().Bool("print-config", false, "output active configuration")
```
Piccirello, line 146: I also considered naming this `--debug-config`, which would mean that the flag would only work when specified with `--debug` (i.e. `--debug --debug-config`). Lmk your thoughts.

bvallelunga: I like this

bvallelunga (approved):

PR #102, opened 2020-06-17
Ignore interrupts during 'doppler run' on Windows

Piccirello:
On Windows, sending an interrupt (e.g. Ctrl+C) would result in the parent process terminating before the user could respond to the child process's "Terminate batch job?" prompt.

Windows 64-bit build: [doppler.zip](https://github.com/DopplerHQ/cli/files/4796407/doppler.zip)

### DIFF
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/signal"
 	"os/user"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


pkg/utils/util.go
```diff
@@ -131,7 +125,29 @@ func RunCommandString(command string, env []string, inFile *os.File, outFile *os
...
+
+	// interrupts will be sent to the entire process group (parent and child processes)
+	// catch + ignore the parent's interrupt so that we use the exit code from the child
+	channel := make(chan os.Signal, 1)
+	signal.Notify(channel, os.Interrupt, syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGTERM)
+	go func() {
+		s := <-channel
+		// do nothing; the child process will have also received the signal
+		LogDebug(fmt.Sprintf("Process received %s", s.String()))
+	}()
```
bvallelunga, line 153: Is this a windows thing where both the parent and child receive the signal?

bvallelunga (approved):

pkg/utils/util.go
Piccirello: It's not Windows specific. The child process will be in the same process group as the parent and so it will receive the same signals.

bvallelunga: Hmm how come this isn't an issue for Mac or linux users then? Is Window's doing something different?

Piccirello: Windows's `cmd` displays a nonsense "Terminate batch job" prompt that keeps the child process from dying. Before the user can respond the parent process has already been killed.

Piccirello:
After some undesirable results when testing on linux and macOS, I've reverted this PR to only affect execution of commands on Windows.

bvallelunga (approved):

PR #105, opened 2020-07-09
Add additional checking for required parameters

Piccirello:
This allows us to quickly show an error instead of trying the request (and failing).


bvallelunga (approved):

bvallelunga:
Nice touch!

PR #106, opened 2020-07-09
Warn user during login if scope already contains a token

Piccirello:
This warning is consistent with how we handle other warnings in the CLI. We show a message with yellow warning text, but otherwise don't get in the user's way.

Screen Shot 2020-07-08 at 3 07 14 PM

### DIFF
--- a/pkg/cmd/login.go
+++ b/pkg/cmd/login.go
@@ -43,6 +43,15 @@ var loginCmd = &cobra.Command{
 		copyAuthCode := !utils.GetBoolFlag(cmd, "no-copy")
 		hostname, _ := os.Hostname()
 
+		// warn user if scope already contains a token
+		if prevConfig.Token.Value != "" {
+			prevScope, err1 := filepath.Abs(prevConfig.Token.Scope)
+			newScope, err2 := filepath.Abs(scope)
+			if err1 == nil && err2 == nil && prevScope == newScope {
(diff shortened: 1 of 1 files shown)


pkg/cmd/login.go
```diff
@@ -43,6 +43,15 @@ var loginCmd = &cobra.Command{
 		copyAuthCode := !utils.GetBoolFlag(cmd, "no-copy")
 		hostname, _ := os.Hostname()
 
+		// warn user if scope already contains a token
+		if prevConfig.Token.Value != "" {
+			prevScope, err1 := filepath.Abs(prevConfig.Token.Scope)
+			newScope, err2 := filepath.Abs(scope)
+			if err1 == nil && err2 == nil && prevScope == newScope {
+				utils.LogWarning("This scope already contains a token and will be overridden.")
```
bvallelunga, line 51: can we add a prefix to the message like: `IMPORTANT: This scope already contains a token and will be overridden.`?

bvallelunga (approved):

pkg/cmd/login.go
Piccirello: It already has a yellow `Warning` prefix.

bvallelunga: You are right.

bvallelunga: Merge and ship

PR #110, opened 2020-07-13
Switch default login scope from "*" to "/"

Piccirello:
With this switch we can now more strictly enforce that the provided scope is a path. This change is completely backwards compatible. Any config containing "*" will be rewritten to use "/", which all old CLI versions already support.

### DIFF
--- a/pkg/cmd/enclave_configs_logs.go
+++ b/pkg/cmd/enclave_configs_logs.go
@@ -78,7 +78,6 @@ var configsLogsRollbackCmd = &cobra.Command{
 	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		jsonFlag := utils.OutputJSON
-		silent := utils.GetBoolFlag(cmd, "silent")
 		localConfig := configuration.LocalConfig(cmd)
 
 		utils.RequireValue("token", localConfig.Token.Value)
@@ -96,7 +95,7 @@ var configsLogsRollbackCmd = &cobra.Command{
 			utils.HandleError(err.Unwrap(), err.Message)
 		}
 
-		if !silent {
+		if !utils.Silent {
 			printer.ConfigLog(configLog, jsonFlag, true)
 		}
 	},
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -40,6 +40,9 @@ var UserConfigDir string
 // UserConfigFile (e.g. /home/user/doppler/.doppler.yaml)
 var UserConfigFile string
 
+// Scope to use for config file
+var Scope = "."
+
 var configFileName = ".doppler.yaml"
 var configContents models.ConfigFile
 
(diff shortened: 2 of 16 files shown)


pkg/configuration/config.go
```diff
@@ -239,10 +248,9 @@ func AllConfigs() map[string]models.FileScopedOptions {
 
 // Set properties on a scoped config
 func Set(scope string, options map[string]string) {
-	scope = NormalizeScope(scope)
 	var err error
-	if scope, err = utils.ParsePath(scope); err != nil {
-		utils.HandleError(err)
+	if scope, err = NormalizeScope(scope); err != nil {
+		utils.HandleError(err, "Invalid scope")
```
bvallelunga, line 253: Can we provide more human errors? Like `The provided scope does not exist on the file system`.

pkg/configuration/migration.go
```diff
@@ -47,8 +47,11 @@ func convertOldConfig(oldConfig map[string]oldConfig) models.ConfigFile {
 	config := map[string]models.FileScopedOptions{}
 
 	for key, val := range oldConfig {
-		key = NormalizeScope(key)
-		config[key] = models.FileScopedOptions{EnclaveProject: val.Pipeline, EnclaveConfig: val.Environment, Token: val.Key}
+		var err error
+		// skip items that fail to parse
+		if key, err = NormalizeScope(key); err == nil {
+			config[key] = models.FileScopedOptions{EnclaveProject: val.Pipeline, EnclaveConfig: val.Environment, Token: val.Key}
```
bvallelunga, line 53: Can at some point we completely migrate off of the legacy terms like `Pipeline` and `Environment`?

bvallelunga (approved):

pkg/configuration/migration.go
Piccirello: The CLI has never used terms like those. This is the migration script to handle users switching from the Node CLI.

pkg/configuration/config.go
Piccirello: This function doesn't check whether the path exists, it just checks if the path contains all valid characters.

Piccirello:
@rgmvisser FYI if you want to test this code, a macOS build is automatically included as a build artifact of the "Release" GitHub Action.

pkg/cmd/enclave_configs_logs.go
```diff
@@ -96,7 +95,7 @@ var configsLogsRollbackCmd = &cobra.Command{
 			utils.HandleError(err.Unwrap(), err.Message)
 		}
 
-		if !silent {
+		if !utils.Silent {
```
rgmvisser, line 98: No really related to this PR i think, but instead having to wrap this around everywhere we print, it would be easier if we just have this in the `printer.ConfigLog(` method or a wrapper around this so that we don't need this if statements every/potentially forget about it.

rgmvisser (approved):

pkg/cmd/enclave_configs_logs.go
Piccirello: I considered making that change when updating all of these. Then I realized that we sometimes want our print functions to print regardless of `utils.Silent`, so I opted not to change for now.

PR #111, opened 2020-07-13
 Prompt user for new path on login conflict 

Piccirello:
Screen Shot 2020-07-13 at 1 16 09 PM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/DopplerHQ/cli
 go 1.13
 
 require (
-	github.com/AlecAivazis/survey/v2 v2.0.4
+	github.com/AlecAivazis/survey/v2 v2.0.8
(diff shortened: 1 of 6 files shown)


bvallelunga (approved):

rgmvisser (changes requested):
Talked offline:
Let's make a menu where the user can pick current folder (scope) or pick overwrite

Piccirello:
@rgmvisser  New version has 3 different prompts, based on the user's situation:

Using default global scope:
Screen Shot 2020-07-13 at 3 16 12 PM

Using non-global scope:
Screen Shot 2020-07-13 at 3 16 56 PM

Using current directory as scope:
Screen Shot 2020-07-13 at 3 17 13 PM

rgmvisser (approved):

bvallelunga (approved):

PR #113, opened 2020-07-25
Log failed http request retry attempts

Piccirello:


bvallelunga:
I thought we were going to do some smart approach where we time the total time and if goes above X seconds we show a message.

In that model we only do 1 print which I think is a cleaner experience.

Piccirello:
I don't know if I agree with the approach of printing once. A fetch could take 50 seconds if exhausting all retries and all timeouts. I've modified this PR to print all retry attempts after 10 seconds.

bvallelunga (approved):

PR #115, opened 2020-07-28
Update status page url

Piccirello:


pkg/cmd/open.go
```diff
@@ -45,7 +45,7 @@ var openStatusCmd = &cobra.Command{
 	Short: "open the Doppler status page",
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
-		err := open.Run("https://status.doppler.com")
+		err := open.Run("https://www.dopplerstatus.com")
```
bvallelunga, line 48:
```suggestion
		err := open.Run("https://dopplerstatus.com")
```

bvallelunga (approved):

pkg/cmd/open.go
Piccirello: It 301 redirects to `www`, might as well save users the redirect.

PR #116, opened 2020-08-10
Add install.sh option to download binary w/o installing

Piccirello:
This allows the install script to be used by users lacking the required perms to install to `/usr/bin/local` and/or `/usr/bin`.


rgmvisser (approved):
Looks good to me. Can you in the future add a description in the PR why we are making these changes?

PR #117, opened 2020-08-11
Add `doppler update` support for all mac and linux installs

Piccirello:
We currently support `doppler update` on macOS via brew. We now drop support for brew updates and instead use our install.sh script for installing updates.

### DIFF
--- a/pkg/cmd/update.go
+++ b/pkg/cmd/update.go
@@ -17,9 +17,8 @@ package cmd
 
 import (
 	"fmt"
-	"os/exec"
-	"regexp"
 
+	"github.com/DopplerHQ/cli/pkg/controllers"
 	"github.com/DopplerHQ/cli/pkg/utils"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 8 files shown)


pkg/cmd/update.go
```diff
@@ -29,50 +35,85 @@ var updateCmd = &cobra.Command{
...
+	if !apiErr.IsNil() {
+		utils.HandleError(apiErr.Unwrap(), apiErr.Message)
+	}
+
+	tmpFile, err := utils.WriteTempFile("install.sh", script, 0555)
+	// clean up temp file once we're done with it
+	defer os.Remove(tmpFile)
+
+	utils.LogDebug("Executing install script")
+	command := []string{tmpFile, "--debug"}
```
rgmvisser: Should we run this without --debug?

rgmvisser (approved):
Looks good to me, 1 small Q

pkg/cmd/update.go
Piccirello: We run with debug so that we can capture additional output. The output is only displayed when the user runs `doppler update` with the `--debug` flag.

Piccirello:
Did a minor refactor, mostly just moving logic around.

PR #119, opened 2020-08-11
Revert "Support CLI updates on Windows via scoop"

Piccirello:
Reverts DopplerHQ/cli#118

This worked great in development, but failed miserably once released. Scoop does not currently support upgrading a running application. Based on our current architecture, the CLI is updating itself so it must be running.

Screen Shot 2020-08-11 at 3 29 50 PM


rgmvisser (approved):
Bummer!

PR #120, opened 2020-08-31
Increase max number of idle connections per host

Piccirello:
This is an attempt to fix the "connect: cannot assign requested address" error some users are seeing. (See internal Intercom message)

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -119,17 +119,22 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 	}
 	req.URL.RawQuery = query.Encode()
 
-	// set timeout and tls config
 	client := &http.Client{}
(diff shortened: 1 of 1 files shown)


pkg/http/http.go
```diff
@@ -119,17 +119,22 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
...
 	req.URL.RawQuery = query.Encode()
 
-	// set timeout and tls config
 	client := &http.Client{}
+	// set http timeout
 	if UseTimeout {
 		client.Timeout = TimeoutDuration
 	}
+
+	// try to prevent "connect: cannot assign requested address" errors
```
rgmvisser, line 128: Can you explain a bit more what the context is around this?

rgmvisser (approved):

PR #121, opened 2020-09-01
Add flag to preserve existing env vars during `doppler run`

Piccirello:
This flag alters `doppler run` behavior such that a Doppler secret is only used if the secret isn't already defined in the current environment.

Closes DPLR-378.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -69,6 +69,7 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 		fallbackOnly := utils.GetBoolFlag(cmd, "fallback-only")
 		exitOnWriteFailure := !utils.GetBoolFlag(cmd, "no-exit-on-write-failure")
 		silentExit := utils.GetBoolFlag(cmd, "silent-exit")
+		preserveEnv := utils.GetBoolFlag(cmd, "preserve-env")
 		localConfig := configuration.LocalConfig(cmd)
 
 		utils.RequireValue("token", localConfig.Token.Value)
@@ -116,18 +117,38 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 
 		secrets := getSecrets(localConfig, enableFallback, fallbackPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase)
 
+		if preserveEnv {
+			utils.LogWarning("Ignoring Doppler secrets already defined in the environment due to --preserve-env flag")
+		}
+
 		env := os.Environ()
+		existingEnvKeys := map[string]bool{}
+		for _, envVar := range env {
+			// key=value format
+			parts := strings.SplitN(envVar, "=", 2)
+			key := parts[0]
+			existingEnvKeys[key] = true
+		}
+
 		excludedKeys := []string{"PATH", "PS1", "HOME"}
 		for name, value := range secrets {
-			addKey := true
+			useSecret := true
 			for _, excludedKey := range excludedKeys {
 				if excludedKey == name {
-					addKey = false
+					useSecret = false
 					break
 				}
 			}
 
-			if addKey {
+			if useSecret && preserveEnv {
+				// skip secret if environment already contains variable w/ same name
+				if existingEnvKeys[name] == true {
+					utils.LogDebug(fmt.Sprintf("Ignoring Doppler secret %s", name))
+					useSecret = false
+				}
+			}
+
+			if useSecret {
 				env = append(env, fmt.Sprintf("%s=%s", name, value))
 			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/cmd/run.go
```diff
@@ -127,6 +132,18 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 				}
 			}
 
+			if addKey && preserveEnv {
+				// skip key if environment variables already contain this key
+				envKey := fmt.Sprintf("%s=", name)
+				for _, envVariable := range env {
+					if strings.HasPrefix(envVariable, envKey) {
```
bvallelunga: Why not check if the strings are equal?

bvallelunga (approved):

pkg/cmd/run.go
```diff
@@ -116,6 +117,10 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 
 		secrets := getSecrets(localConfig, enableFallback, fallbackPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase)
 
+		if preserveEnv {
+			utils.LogWarning("Ignoring any secrets already defined in the environment due to --preserve-env flag")
+		}
+
 		env := os.Environ()
 		excludedKeys := []string{"PATH", "PS1", "HOME"}
```
bvallelunga, line 133: Unrelated but why do we exclude these secrets?

pkg/cmd/run.go
Piccirello: We don't care if the variables have the same value, just that they share the same name.

bvallelunga: Oh i see

pkg/cmd/run.go
```diff
@@ -127,6 +132,18 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 				}
 			}
 
+			if addKey && preserveEnv {
+				// skip key if environment variables already contain this key
+				envKey := fmt.Sprintf("%s=", name)
```
bvallelunga: Could this introduce a bug where if you are trying to exclude `E` then `APPLE` will also be excluded?

bvallelunga (changes requested):

pkg/cmd/run.go
bvallelunga: Can we instead get just the name of the env var and compare it to the secret name?

Piccirello: I don't see how it could

bvallelunga: Ohh because you are doing a hasPrefix, it safeguards against that case.

bvallelunga (approved):

pkg/cmd/run.go
Piccirello: IIRC these were copied over from the node cli

pkg/cmd/run.go
rgmvisser: is the prefix because the var `envVariable` now is:
`TOKEN=1234324545345` ?
If that is the case can you strip out the `envVariable` the same way we do `envKey`? It's bit confusing otherwise

pkg/cmd/run.go
```diff
@@ -127,6 +132,18 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
+				for _, envVariable := range env {
+					if strings.HasPrefix(envVariable, envKey) {
+						utils.LogDebug(fmt.Sprintf("Ignoring Doppler secret %s", name))
```
rgmvisser: Should this be `envKey`?

Piccirello: Given a doppler secret `API_KEY`, `name` is `API_KEY` and `envKey` is `API_KEY=`

pkg/cmd/run.go
Piccirello: I've updated the implementation so that it's more straightforward

pkg/cmd/run.go
```diff
@@ -116,18 +117,40 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 
 		secrets := getSecrets(localConfig, enableFallback, fallbackPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase)
 
+		if preserveEnv {
+			utils.LogWarning("Ignoring Doppler secrets already defined in the environment due to --preserve-env flag")
+		}
+
 		env := os.Environ()
+		existingEnvKeys := []string{}
```
rgmvisser: Very small nit, but we should use a map so that we don't have to loop over this array every time

rgmvisser (approved):

pkg/cmd/run.go
Piccirello: Whoops, good catch

PR #125, opened 2020-09-10
Deprecate Enclave from the CLI

Piccirello:
This PR deprecates Enclave from the CLI. It is entirely backwards compatible and will be a minor release.

- Deprecate all Enclave commands (e.g. `doppler enclave configs logs` is now `doppler configs logs`)
- Deprecate `--no-update-enclave-config` flag and replace with `--no-update config-options`
- Accept user input for `enclave.config` and `enclave.project` config options as `config` and `project`, respectively
- Display `enclave.config` and `enclave.project` as `config` and `project` in all tables (json is unmodified for backwards compatibility)

macOS build: [doppler.tar.gz](https://github.com/DopplerHQ/cli/files/5217195/doppler.tar.gz)


Piccirello:
I need to figure out why these builds are failing. They're working fine locally.

rgmvisser (commented):
This is a very big one, it's a bit hard to review as a lot of logic is moved around (I assume it's all copy pasta). Is there an easy way I can test this locally?

rgmvisser:
Btw, in general it looks good, but just wanna test a couple edge cases

Piccirello:
> This is a very big one, it's a bit hard to review as a lot of logic is moved around (I assume it's all copy pasta). Is there an easy way I can test this locally?

The only logic changes are in the last two commits (4b0372109ff5af8e213fd6df794073c1bbdca99b and dea92a977b8f75ac16fc1776233772a4873a58cf). Everything else is just moving code around so it can be reused.

> Is there an easy way I can test this locally?

Typically the "release" GitHub Action uploads a macOS build as an artifact. In this case the build is failing so no artifact. Here's my latest local build: [doppler.zip](https://github.com/DopplerHQ/cli/files/5210834/doppler.zip)

rgmvisser (approved):
Looks good to me, 1 small nit:

if I just run doppler it still shows this in the command list, can we get rid of that?
`enclave      Manage Enclave`

Piccirello:
> if I just run doppler it still shows this in the command list, can we get rid of that?

Great idea, just added

bvallelunga:
CLI looks great outside of the `enclave.config` and `enclave.project` showing in the table. I tested setup, projects list, secrets list, and run.

Piccirello:
I've pushed up a couple commits that address the `enclave.config`/`enclave.project` issue. We now drop the `enclave.` prefix when printing config values in a table (e.g. `enclave.config` is printed as `config`). JSON output is unchanged for backwards compatibility. We also now accept `config` and `project` in `doppler configure` commands (e.g. `doppler configure set config=dev`). These changes don't modify the config file in any way.

Updated build: [doppler.tar.gz](https://github.com/DopplerHQ/cli/files/5217195/doppler.tar.gz)

rgmvisser (approved):
@Piccirello looks good to me, can we add a task/todos in the code that we can easily remove this when we upgrade to a new major version? There is a lot of ugly stuff now in there for backward compatibility :P

Piccirello:
Required work for CLI v4 is documented in DPLR-435, I'll add a reference to it in the codebase.

PR #126, opened 2020-09-11
Improve HTTP edge case handling

Piccirello:
-  Handle edge case where response body may not be closed 
-  Ensure HTTP connections are closed after each request 
-  Properly handle errors with non-json response

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -119,6 +119,9 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 	}
 	req.URL.RawQuery = query.Encode()
 
+	// close the connection after reading the response, to help prevent socket exhaustion
+	req.Close = true
+
 	client := &http.Client{}
 	// set http timeout
 	if UseTimeout {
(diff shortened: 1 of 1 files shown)


pkg/http/http.go
```diff
@@ -144,6 +147,10 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 	requestErr := retry(5, 100*time.Millisecond, func() error {
 		resp, err := client.Do(req)
 		if err != nil {
+			if resp != nil {
+				defer response.Body.Close()
```
rgmvisser: `response` is still nil here, should this be `resp`?

Piccirello: Great catch

pkg/http/http.go
```diff
@@ -144,6 +147,10 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 	requestErr := retry(5, 100*time.Millisecond, func() error {
 		resp, err := client.Do(req)
 		if err != nil {
+			if resp != nil {
+				defer resp.Body.Close()
+			}
```
rgmvisser, line 152: Small nit, but can we just take this outside of the `if err != nil {` in that case the 2nd defer is not needed anymore (and less likely to introduce a bug in the future)

rgmvisser (approved):

pkg/http/http.go
Piccirello: I don't believe we can due to this code executing inside a function. From my understanding, once the function returns the deferred functions will execute, meaning the response body will be closed before we have a chance to read it below.

rgmvisser: Yup you are right 👍

PR #127, opened 2020-09-16
Forward all signals to subprocess during `doppler run`

Piccirello:


Piccirello:
Testing looks good on macOS, Windows, and linux.

bvallelunga (approved):

PR #130, opened 2020-09-26
Use Doppler v3 APIs

Piccirello:
This PR implements support for the Doppler v3 APIs. This also means that the CLI no longer requires specifying a project or config when using a service token.

The default fallback file path and passphrase have changed in this release. To maintain backwards compatibility, we write both the old file and the new file. Additionally, we support reading from both fallback files (with precedence given to the new file).

### DIFF
--- a/pkg/cmd/configs.go
+++ b/pkg/cmd/configs.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
@@ -160,7 +161,12 @@ func deleteConfigs(cmd *cobra.Command, args []string) {
 	}
 	utils.RequireValue("config", config)
 
-	if yes || utils.ConfirmationPrompt("Delete config "+config, false) {
+	prompt := "Delete config"
+	if config != "" {
+		prompt = fmt.Sprintf("%s %s", prompt, config)
+	}
+
+	if yes || utils.ConfirmationPrompt(prompt, false) {
 		err := http.DeleteConfig(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, config)
 		if !err.IsNil() {
 			utils.HandleError(err.Unwrap(), err.Message)
@@ -216,7 +222,12 @@ func lockConfigs(cmd *cobra.Command, args []string) {
 	}
 	utils.RequireValue("config", config)
 
-	if yes || utils.ConfirmationPrompt("Lock config "+config, false) {
+	prompt := "Lock config"
+	if config != "" {
+		prompt = fmt.Sprintf("%s %s", prompt, config)
+	}
+
+	if yes || utils.ConfirmationPrompt(prompt, false) {
 		configInfo, err := http.LockConfig(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, config)
 		if !err.IsNil() {
 			utils.HandleError(err.Unwrap(), err.Message)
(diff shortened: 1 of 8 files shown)


Piccirello:
Salus error fixed in #131

bvallelunga (approved):

pkg/cmd/configs.go
```diff
@@ -152,15 +149,18 @@ func deleteConfigs(cmd *cobra.Command, args []string) {
...
 	}
-	utils.RequireValue("config", config)
 
-	if yes || utils.ConfirmationPrompt("Delete config "+config, false) {
+	prompt := "Delete config"
+	if config != "" {
+		prompt = fmt.Sprintf("%s %s", prompt, config)
+	}
+
+	if yes || utils.ConfirmationPrompt(prompt, false) {
```
rgmvisser, line 163: This is always true right? What is the logic behind this?

pkg/cmd/run.go
```diff
@@ -73,18 +73,22 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
 		if cmd.Flags().Changed("fallback") {
 			var err error
 			fallbackPath, err = utils.GetFilePath(cmd.Flag("fallback").Value.String())
 			if err != nil {
 				utils.HandleError(err, "Unable to parse --fallback flag")
 			}
 		} else {
-			fallbackPath = defaultFallbackFile(localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+			fallbackPath = defaultFallbackFile(localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+			// TODO remove this when releasing CLI v4
```
rgmvisser: Can you add a task?

pkg/cmd/run.go
```diff
@@ -285,6 +302,19 @@ func fetchSecrets(localConfig models.ScopedOptions, enableFallback bool, fallbac
 				utils.LogError(err)
 			}
 		}
+
+		// TODO remove this when releasing CLI v4
```
rgmvisser: Task pls :)

pkg/cmd/run.go
```diff
@@ -73,18 +73,22 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
 			if err != nil {
 				utils.HandleError(err, "Unable to parse --fallback flag")
 			}
 		} else {
-			fallbackPath = defaultFallbackFile(localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+			fallbackPath = defaultFallbackFile(localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+			// TODO remove this when releasing CLI v4
+			if localConfig.EnclaveProject.Value != "" && localConfig.EnclaveConfig.Value != "" {
+				// save to old path to maintain backwards compatibility
+				legacyFallbackPath = legacyFallbackFile(localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
```
rgmvisser, line 90: I assume this is on the writing side? Why not read to the new path for these as well?

rgmvisser (changes requested):
Couple minor comments and a small Q

pkg/cmd/configs.go
Piccirello: This is only true when the user specifies the `--yes` flag or they answer `yes` to the prompt

pkg/cmd/run.go
Piccirello: Updated the comment

pkg/cmd/run.go
Piccirello: We support reading from the legacy path as well in `readFallbackFile()`

pkg/cmd/configs.go
rgmvisser: Ah I see, somehow my hungover brain was thinking that yes in go is true :P

rgmvisser (approved):

PR #132, opened 2020-09-26
Allow updating a project's name without the description

Piccirello:
We previously required both the name and the description to be updated.

### DIFF
--- a/pkg/http/api.go
+++ b/pkg/http/api.go
@@ -306,9 +306,14 @@ func CreateProject(host string, verifyTLS bool, apiKey string, name string, desc
 	return projectInfo, Error{}
 }
 
-// UpdateProject update a project
-func UpdateProject(host string, verifyTLS bool, apiKey string, project string, name string, description string) (models.ProjectInfo, Error) {
-	postBody := map[string]string{"name": name, "description": description}
+// UpdateProject update a project's name and (optional) description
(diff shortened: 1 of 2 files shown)


pkg/http/api.go
```diff
@@ -306,9 +306,14 @@ func CreateProject(host string, verifyTLS bool, apiKey string, name string, desc
...
 }
 
-// UpdateProject update a project
-func UpdateProject(host string, verifyTLS bool, apiKey string, project string, name string, description string) (models.ProjectInfo, Error) {
-	postBody := map[string]string{"name": name, "description": description}
+// UpdateProject update a project's name and (optional) description
+func UpdateProject(host string, verifyTLS bool, apiKey string, project string, name string, description ...string) (models.ProjectInfo, Error) {
+	postBody := map[string]string{"name": name}
+	if len(description) > 0 {
+		desc := description[0]
```
rgmvisser, line 313: super nit: `postBody["description"] := description[0]`

rgmvisser: Why is the description an array btw?

rgmvisser (approved):

pkg/http/api.go
Piccirello: We do this so that the description can be optional. We need to support `""` as a valid value, so we can't use that to signal whether a description was specified.

PR #138, opened 2020-10-10
Use home dir as temp directory

Piccirello:
This prevents temp directories from being created in /tmp/, which is world writable. This fixes a potential race condition where the downloaded CLI install script and CLI binary could be replaced by an attacker before execution.

Tested the new `mktemp` command on macOS, Ubuntu, alpine, RHEL, CentOS, and FreeBSD.

Closes DPLR-502.


Piccirello:
For more info see DPLR-502.

bvallelunga (approved):

PR #139, opened 2020-10-13
Reap zombie processes in Docker using tini

Piccirello:
This prevents the Docker zombie reaping problem[0] that's caused by pid 1 (in this case the CLI) not sufficiently cleaning up zombie processes. Without adding an init process to the entrypoint, another solution is using the `--init` flag on `docker run`.

[0] https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

### DIFF
--- a/docker/alpine
+++ b/docker/alpine
@@ -1,3 +1,4 @@
 FROM alpine
+RUN apk add --no-cache tini
 COPY doppler /bin/doppler
-ENTRYPOINT ["/bin/doppler"]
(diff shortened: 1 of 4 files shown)


bvallelunga (approved):

Piccirello:
Added note to [CLI docs](https://docs.doppler.com/docs/enclave-installation-docker#running-the-image) to specify `--init` when using a custom Docker image.

PR #140, opened 2020-10-13
Add fallback file support to the 'secrets download' command

Piccirello:
Fallback files only work when used with the default JSON format (.env is not supported).

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -218,6 +218,7 @@ var runCleanCmd = &cobra.Command{
 	},
 }
 
+// fetchSecrets fetches secrets, including all reading and writing of fallback files
 func fetchSecrets(localConfig models.ScopedOptions, enableFallback bool, fallbackPath string, legacyFallbackPath string, fallbackReadonly bool, fallbackOnly bool, exitOnWriteFailure bool, passphrase string) map[string]string {
 	if fallbackOnly {
 		if !enableFallback {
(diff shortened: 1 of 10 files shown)


Piccirello:
Added a bunch of E2E tests for this feature, as well as a hacked-together test runner. I think this hacky test runner is sufficient for now without investing a ton more time.

pkg/cmd/run.go
```diff
@@ -396,6 +386,19 @@ func defaultFallbackFile(token string, project string, config string) string {
...
+func getPassphrase(cmd *cobra.Command, flag string, config models.ScopedOptions) string {
+	if cmd.Flags().Changed(flag) {
+		return cmd.Flag(flag).Value.String()
+	}
+
+	if config.EnclaveProject.Value != "" && config.EnclaveConfig.Value != "" {
+		return fmt.Sprintf("%s:%s:%s", config.Token.Value, config.EnclaveProject.Value, config.EnclaveConfig.Value)
+	}
+
+	return config.Token.Value
```
rgmvisser, line 400: Not sure if this matters, but should this be `fmt.Sprintf("%s", localConfig.Token.Value)`?

pkg/cmd/secrets.go
```diff
@@ -218,9 +224,40 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+		enableFallback = false
+		flags := []string{"fallback", "fallback-only", "fallback-readonly", "no-exit-on-write-failure"}
+		for _, flag := range flags {
+			if cmd.Flags().Changed(flag) {
+				utils.LogWarning(fmt.Sprintf("--%s has no effect when format is %s", flag, format))
+			}
+		}
+
+		var apiError http.Error
+		body, apiError = http.DownloadSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, format == "json")
```
rgmvisser: Replace `format == "json"` with false?

pkg/cmd/secrets.go
```diff
@@ -218,9 +224,40 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+	}
+
+	var body []byte
+	if format == "json" {
+		fallbackPath := ""
+		legacyFallbackPath := ""
+		if enableFallback {
+			fallbackPath, legacyFallbackPath = initFallbackDir(cmd, localConfig, exitOnWriteFailure)
+		}
+		secrets := fetchSecrets(localConfig, enableFallback, fallbackPath, legacyFallbackPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, fallbackPassphrase)
```
rgmvisser, line 239: Can you add some comments around here on what this function exactly does? Since the fallback arguments are passed in I feel like it is doing more than just fetching secrets?

tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -euo pipefail
```
rgmvisser, line 3: What does this do? (i am a bash n00b haha) Maybe add a comment?

rgmvisser (dismissed):
Couple small questions, but yeeesss tests

pkg/cmd/run.go
Piccirello: Those two statements are equivalent, so I went with the simpler one

tests.sh
Piccirello: This sets bash's error handling. Specifically:

`-e`: Any line failing causes the whole script to fail
`-u`: Using an undeclared variable raises an error
`-o pipefail`: Respect the exit code of any command that fails within a piped expression

Screenshot of `pipefail` in action. You'll notice that initially, `false | echo ok` uses the exit code of `echo ok`, which is 0. After calling `set -o pipefail` it now respects the exit code of `false`, which is 1.
Screen Shot 2020-10-14 at 9 46 33 PM

rgmvisser: It all makes sense now :)

pkg/cmd/secrets.go
Piccirello: Adding godoc

PR #142, opened 2020-10-15
Fix: forward all signals to subprocess

Piccirello:
When running with a TTY, user-generated signals (like SIGINT/Ctrl+C) are sent to the entire process group of the foreground process. The CLI and its child process run in the same process group, so if the CLI forwards the signal, the child process will receive it twice. In all other scenarios, we want to forward the signal to the child process, otherwise it won't receive it at all.

Fixes: 
- calling `docker stop` on detached docker daemon is ignored when using `doppler run`

Closes DPLR-716

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -31,6 +31,7 @@ import (
 	"github.com/DopplerHQ/cli/pkg/http"
 	"github.com/DopplerHQ/cli/pkg/models"
 	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/mattn/go-isatty"
 	"github.com/spf13/cobra"
 	"gopkg.in/gookit/color.v1"
 )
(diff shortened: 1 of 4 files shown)


rgmvisser (dismissed):

pkg/cmd/run.go
```diff
@@ -539,6 +540,7 @@ func init() {
 	runCmd.Flags().Bool("fallback-readonly", false, "disable modifying the fallback file. secrets can still be read from the file.")
 	runCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
 	runCmd.Flags().Bool("no-exit-on-write-failure", false, "do not exit if unable to write the fallback file")
+	runCmd.Flags().Bool("forward-signals", false, "forward signals to the child process, regardless of whether there's a TTY allocated")
```
rgmvisser: Should this be true by default?

ryan-blunden (dismissed):

pkg/cmd/run.go
Piccirello: I updated the default to reflect the calculated value, and also made the flag's description a bit more clear.

rgmvisser (approved):

PR #144, opened 2020-10-16
Support using fallback file as secrets cache

Piccirello:
Currently, on successful secrets fetch, the CLI encrypts the response and saves it as a fallback file. With this PR, we now additionally save the response ETag to a separate metadata file.

On subsequent fetches, we read the ETag from the metadata file and send it in the request's `If-None-Match` header. If the server confirms that this is still the latest hash, it sends back a 304. The CLI responds to this 304 by reading the fallback file. This results in a speed improvement due to the API performing a fast hash lookup, rather than a slower token reveal.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/controllers"
 	"github.com/DopplerHQ/cli/pkg/crypto"
 	"github.com/DopplerHQ/cli/pkg/http"
 	"github.com/DopplerHQ/cli/pkg/models"
@@ -65,6 +66,7 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		enableFallback := !utils.GetBoolFlag(cmd, "no-fallback")
+		enableCache := enableFallback && !utils.GetBoolFlag(cmd, "no-cache")
 		fallbackReadonly := utils.GetBoolFlag(cmd, "fallback-readonly")
 		fallbackOnly := utils.GetBoolFlag(cmd, "fallback-only")
 		exitOnWriteFailure := !utils.GetBoolFlag(cmd, "no-exit-on-write-failure")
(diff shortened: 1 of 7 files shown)


pkg/cmd/run.go
```diff
@@ -75,9 +77,13 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 
 		fallbackPath := ""
 		legacyFallbackPath := ""
+		metadataPath := ""
 		if enableFallback {
 			fallbackPath, legacyFallbackPath = initFallbackDir(cmd, localConfig, exitOnWriteFailure)
 		}
+		if enableCache {
+			metadataPath = controllers.MetadataFilePath(localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
```
rgmvisser, line 85: Would this error if `localConfig.EnclaveProject` is null?

pkg/cmd/run.go
```diff
@@ -278,6 +298,14 @@ func fetchSecrets(localConfig models.ScopedOptions, enableFallback bool, fallbac
 				}
 			}
 		}
+
+		if enableCache {
```
rgmvisser, line 312: Not sure if this is needed, but you are doing it in other places:
`if enableCache && metadataPath != "" {`

rgmvisser (dismissed):
Very excited about this!

pkg/cmd/run.go
Piccirello: `MetadataFilePath` handles the case when `EnclaveProject` and `EnclaveConfig` are empty (`""`)

pkg/cmd/run.go
Piccirello: Great catch

rgmvisser (dismissed):

Piccirello:
I pushed up some minor changes to the error handling logic

pkg/cmd/run.go
```diff
@@ -65,6 +66,7 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		enableFallback := !utils.GetBoolFlag(cmd, "no-fallback")
+		enableCache := enableFallback && !utils.GetBoolFlag(cmd, "no-cache")
```
rgmvisser, line 69: Just a thought, but now this variable sounds like this not enabled by default, where I think it is by default enabled right?
Should we name the variables so they make sense with the default?

rgmvisser (approved):
Looks good, just 1 question on variable naming

pkg/cmd/run.go
Piccirello: I generally prefer to use variables that indicate whether something is enabled, rather than the reverse, regardless of the default value. We do the same with `enableFallback` (which is enabled by default). This is just personal preference though.

PR #147, opened 2020-10-16
Display CLI changelog with 'doppler changelog' command

Piccirello:
This pulls the changelog from cli.doppler.com/changes, which is a Cloudflare Worker that pulls it from GitHub.

Screen Shot 2020-10-16 at 2 29 30 AM

Most recent changes displayed after updating:
Screen Shot 2020-10-18 at 11 57 43 AM


pkg/cmd/update.go
```diff
@@ -60,6 +61,13 @@ func installCLIUpdate() {
 
 	if wasUpdated {
 		utils.Log(fmt.Sprintf("Installed CLI %s", installedVersion))
+		utils.Log("\nWhat's new:")
```
rgmvisser: We should move this inside the if statement in case we don't get a result.

rgmvisser (dismissed):
This is great! A small addition, but really nice. Maybe we can also print something like:
`run 'doppler changelog' to see all our latest changes`?

rgmvisser (changes requested):
back in your queue

Piccirello:
Screen Shot 2020-10-27 at 6 09 25 PM

rgmvisser (approved):

PR #148, opened 2020-10-17
Save auth tokens to the OS keyring

Piccirello:
This prevents tokens from being written to the filesystem. Supported on macOS, Windows, and Linux.

### DIFF
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/DopplerHQ/cli/pkg/controllers"
 	"github.com/DopplerHQ/cli/pkg/models"
 	"github.com/DopplerHQ/cli/pkg/utils"
 	"github.com/spf13/cobra"
@@ -143,6 +144,16 @@ func Get(scope string) models.ScopedOptions {
 		}
 	}
 
+	if controllers.IsKeyringSecret(scopedConfig.Token.Value) {
+		utils.LogDebug(fmt.Sprintf("Retrieving %s from system keyring", models.ConfigToken.String()))
+		token, err := controllers.GetKeyring(scopedConfig.Token.Value)
+		if !err.IsNil() {
+			utils.HandleError(err.Unwrap(), err.Message)
+		}
+
+		scopedConfig.Token.Value = token
+	}
+
 	return scopedConfig
 }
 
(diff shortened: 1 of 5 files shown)


pkg/configuration/config.go
```diff
@@ -266,12 +295,44 @@ func Set(scope string, options map[string]string) {
...
-		config := configContents.Scoped[normalizedScope]
+		if key == models.ConfigToken.String() {
+			utils.LogDebug(fmt.Sprintf("Saving %s to system keyring", key))
+			uuid, err := uuid.NewRandom()
+			if err != nil {
+				utils.HandleError(err, "Unable to generate random UUID")
+			}
+			id := controllers.GenerateKeyringID(uuid.String())
+
+			if controllerError := controllers.SetKeyring(id, value); !controllerError.IsNil() {
```
rgmvisser, line 312: Can you add some comment to the different path here? It looks like this is some backward compatibility logic, but not sure

pkg/configuration/config.go
```diff
@@ -266,12 +295,44 @@ func Set(scope string, options map[string]string) {
...
+				err := controllerError.Unwrap()
+				if err == keyring.ErrUnsupportedPlatform {
+					utils.LogDebug("Your OS does not support keyring")
+				} else {
+					utils.LogDebugError(err)
+					utils.LogDebug("Unable to access system keyring for secure storage")
+				}
```
rgmvisser:
```
utils.LogDebugError(controllerError.Unwrap())
utils.LogDebug(controllerError.Message)
```
?

rgmvisser (dismissed):
Looks good in general, couple of clarification comments and nits.
Just to be sure, if the platform does not support keychain, for instance prod linux server, we just save it plain?

pkg/configuration/config.go
Piccirello: Can you clarify what you mean here, I don't follow

rgmvisser: Nvm, ignore

Piccirello:
>  Just to be sure, if the platform does not support keychain, for instance prod linux server, we just save it plain?

We discussed this in person, but for posterity: exactly!

pkg/configuration/config.go
Piccirello: Updated, good catch

rgmvisser (dismissed):

Piccirello:
@rgmvisser pushed up a tiny change to make the keyring prefix check less error prone.

rgmvisser (approved):

PR #151, opened 2020-10-28
Use V3 APIs for CLI auth

Piccirello:
These were mistakenly never switched over from the V1 APIs.


rgmvisser:
Is this backward compatible?

Piccirello:
It is, the old and new APIs have identical responses.

Piccirello:
PR that this change should have been a part of: https://github.com/DopplerHQ/cli/pull/130

rgmvisser (approved):

PR #158, opened 2020-11-11
Add repo config file functionality DPLR-712

rgmvisser:
You can now have a `doppler.yaml` that can have the project and config specified for the repo

### DIFF
--- a/pkg/cmd/setup.go
+++ b/pkg/cmd/setup.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/controllers"
 	"github.com/DopplerHQ/cli/pkg/http"
 	"github.com/DopplerHQ/cli/pkg/models"
 	"github.com/DopplerHQ/cli/pkg/printer"
@@ -56,6 +57,12 @@ func setup(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	repoConfig, err := controllers.RepoConfig()
+	if !err.IsNil() {
+		utils.Log(err.Message)
+		utils.LogDebugError(err.Unwrap())
+	}
+
 	currentProject := localConfig.EnclaveProject.Value
 	selectedProject := ""
 
@@ -66,6 +73,11 @@ func setup(cmd *cobra.Command, args []string) {
 		utils.Log(valueFromEnvironmentNotice("DOPPLER_PROJECT"))
 		selectedProject = localConfig.EnclaveProject.Value
 	default:
+		if repoConfig.Setup.Project != "" {
+			utils.Log("Reading project from repo config file doppler.yaml")
+			selectedProject = repoConfig.Setup.Project
+			break
+		}
 		projects, httpErr := http.GetProjects(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value)
 		if !httpErr.IsNil() {
 			utils.HandleError(httpErr.Unwrap(), httpErr.Message)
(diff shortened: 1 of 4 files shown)


pkg/controllers/repo_config.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// RepoConfig Reads the configuration file (doppler.yaml) if exists and returns the set configuration
+func RepoConfig() models.RepoConfig {
+
+	if utils.Exists(RepoConfigFile) {
+		yamlFile, err := ioutil.ReadFile(RepoConfigFile) // #nosec G304
+
+		if err != nil {
+			utils.HandleError(err, "Unable to read doppler defaults file")
```
rgmvisser: Should we error here or just return an empty config? If we can't read anything for some reason, it might be better to just gracefully handle it and log a debug statement?

pkg/controllers/repo_config.go
```diff
@@ -0,0 +1,58 @@
...
+// repoConfigFileName (doppler.yaml)
+var repoConfigFileName string
+
+// RepoConfigFile (e.g. ./doppler.yaml)
+var RepoConfigFile string
+
+func init() {
+	repoConfigFileName = "doppler.yaml"
+	RepoConfigFile = filepath.Join("./", repoConfigFileName)
+}
```
Piccirello: If we're setting default values in `init()`, can we just set the values when the variables are declared?

```suggestion
const repoConfigFileName := "doppler.yaml"

// RepoConfigFile (e.g. ./doppler.yaml)
RepoConfigFile := filepath.Join("./", repoConfigFileName)
```

pkg/controllers/repo_config.go
Piccirello:
```suggestion
			utils.HandleError(err, "Unable to read repo config")
```

pkg/controllers/repo_config.go
```diff
@@ -0,0 +1,58 @@
...
+		yamlFile, err := ioutil.ReadFile(RepoConfigFile) // #nosec G304
+
+		if err != nil {
+			utils.HandleError(err, "Unable to read doppler defaults file")
+		}
+
+		var repoConfig models.RepoConfig
+
+		err = yaml.Unmarshal(yamlFile, &repoConfig)
+		if err != nil {
```
Piccirello: nit (mostly stylistic):
```suggestion
		if err := yaml.Unmarshal(yamlFile, &repoConfig); err != nil {
```

pkg/cmd/setup.go
```diff
@@ -66,6 +69,11 @@ func setup(cmd *cobra.Command, args []string) {
 		utils.Log(valueFromEnvironmentNotice("DOPPLER_PROJECT"))
 		selectedProject = localConfig.EnclaveProject.Value
 	default:
+		if repoConfig.Setup.Project != "" {
+			selectedProject = repoConfig.Setup.Project
+			utils.LogDebug("Reading project from doppler.yaml")
```
Piccirello: I think this should always be logged, rather than making it debug-only. This would be similar to line 55 above where we read values out of the environment. To see what I mean, run `DOPPLER_PROJECT=123 doppler setup`.

pkg/controllers/repo_config.go
Piccirello: Good question, I think we should do 3 things:

1. Log something for the user
2. Debug log the error object
3. Return an empty config

It might also be useful if `RepoConfig()` returns the error and allowed the caller to handle it. This would be similar to the `MetadataFile()`. I also think we should debug log when we're reading the file.

pkg/cmd/setup.go
```diff
@@ -66,6 +73,11 @@ func setup(cmd *cobra.Command, args []string) {
 		utils.Log(valueFromEnvironmentNotice("DOPPLER_PROJECT"))
 		selectedProject = localConfig.EnclaveProject.Value
 	default:
+		if repoConfig.Setup.Project != "" {
+			utils.Log("Reading project from doppler.yaml")
```
Piccirello: After testing locally, I think this message should more closely match our error message "Unable to parse doppler repo config file". Otherwise it's not clear to me that these are referring to the same thing.
```suggestion
			utils.Log("Reading project from repo config file doppler.yaml")
```

Piccirello (dismissed):

rgmvisser:
@Piccirello let's chat tomorrow about how to deploy this etc. :)

Piccirello (approved):

PR #159, opened 2020-11-13
Chore: Also allow reading from repo config from .yml

rgmvisser:


pkg/cmd/setup.go
```diff
@@ -74,7 +74,7 @@ func setup(cmd *cobra.Command, args []string) {
 		selectedProject = localConfig.EnclaveProject.Value
 	default:
 		if repoConfig.Setup.Project != "" {
-			utils.Log("Reading project from repo config file doppler.yaml")
+			utils.Log(fmt.Sprintf("Reading project from repo config file %s", repoConfigFile))
```
ryan-blunden, line 77:
```suggestion
			utils.Log(fmt.Sprintf("Reading project from Doppler config file %s", repoConfigFile))
```

pkg/cmd/setup.go
```diff
@@ -103,7 +103,7 @@ func setup(cmd *cobra.Command, args []string) {
 		selectedConfig = localConfig.EnclaveConfig.Value
 	default:
 		if repoConfig.Setup.Config != "" {
-			utils.Log("Reading config from repo config file doppler.yaml")
+			utils.Log(fmt.Sprintf("Reading config from repo config file %s", repoConfigFile))
```
ryan-blunden, line 106:
```suggestion
			utils.Log(fmt.Sprintf("Reading config from Doppler config file %s", repoConfigFile))
```

ryan-blunden (approved):

rgmvisser:
Discussed with @Piccirello, instead of supporting both extension, we will make a changes that we log to the user that they should use .yaml instead of .yml. We prefer to just support 1 extension

PR #161, opened 2020-11-18
Prompt user to use doppler.yaml during setup

Piccirello:
We now prompt the user to determine if we should use the values from doppler.yaml. If they say no, we auto-select the doppler.yaml values in the subsequent project/config selection prompts.

Screen Shot 2020-11-18 at 11 46 23 AM

We also maintain support for the existing `--no-prompt` flag.

Screen Shot 2020-11-18 at 11 56 13 AM

### DIFF
--- a/pkg/cmd/setup.go
+++ b/pkg/cmd/setup.go
@@ -63,6 +63,12 @@ func setup(cmd *cobra.Command, args []string) {
 		utils.LogDebugError(err.Unwrap())
 	}
 
+	// prompt whether to use repoConfig
+	useRepoConfig := true
+	if promptUser && (repoConfig.Setup.Project != "" || repoConfig.Setup.Config != "") {
+		useRepoConfig = utils.ConfirmationPrompt("Use default settings from repo config file (doppler.yaml)?", true)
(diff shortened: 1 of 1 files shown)


pkg/cmd/setup.go
```diff
@@ -63,6 +63,12 @@ func setup(cmd *cobra.Command, args []string) {
 		utils.LogDebugError(err.Unwrap())
 	}
 
+	// prompt whether to use repoConfig
+	useRepoConfig := true
+	if promptUser && (repoConfig.Setup.Project != "" || repoConfig.Setup.Config != "") {
+		useRepoConfig = utils.ConfirmationPrompt("Read values from repo config file (doppler.yaml)?", true)
```
rgmvisser: `Use default setting from repo config file (doppler.yaml)?`?

rgmvisser (dismissed):

pkg/cmd/setup.go
Piccirello: I like that

rgmvisser (approved):

PR #162, opened 2020-11-19
Verify hash of cache file before using

Piccirello:
This handles a previously unconsidered edge case where a valid etag has been cached, but the fallback file points to a different file than what was saved. The fallback file would fail to decrypt and cause the command to exit.

To reproduce locally:
1. Generate metadata file and fallback file: `doppler run --fallback ./fallback.json`
2. Modify fallback file contents: `rm -f ./fallback.json && echo foo > ./fallback.json`
3. Execute again: `doppler run --fallback ./fallback.json`

Closes DPLR-900.


rgmvisser (approved):
I like that we are adding tests now :)

PR #163, opened 2020-11-20
Add 'secrets upload' command for importing files

Piccirello:
`doppler secrets upload <filepath>` will import the contents of the specified file to a config. It currently supports json and env.

I'm toying with calling this `doppler secrets import` instead. What do you guys think?

### DIFF
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


rgmvisser (approved):

ryan-blunden (approved):
Awesome!

ryan-blunden:
@Piccirello are there any limitations format wise, such as no nested JSON or env files requiring values to be quoted?

Piccirello:
Nothing like that, as far as I'm aware. We do support nested JSON. This uses the same code as our "Upload env" feature on our dashboard.

PR #170, opened 2021-01-13
Support downloading secrets in 'env-no-quotes' format

Piccirello:
Closes DPLR-1177


rgmvisser (dismissed):

ryan-blunden:
@Piccirello Looks you need to also update [secrets.go](https://github.com/DopplerHQ/cli/blob/tom/env-no-quotes/pkg/cmd/secrets.go#L398) with the `docker` and now `env-no-quotes` format.

Piccirello:
> @Piccirello Looks you need to also update [secrets.go](https://github.com/DopplerHQ/cli/blob/tom/env-no-quotes/pkg/cmd/secrets.go#L398) with the `docker` and now `env-no-quotes` format.

Good catch, will replace that static list so I can forget about it again :)

rgmvisser (approved):

PR #176, opened 2021-02-15
Add error handling for missing parent dir when using `--configuration`

nmanoogian:
The CLI is designed to create the `UserConfigDir` if it does not already exist. However, when the user specifies `--configuration`, it replaces `UserConfigFile` but not `UserConfigDir` (which is still used in its original form for storing fallback files). To reduce confusion, this change adds a check to make sure that the parent directory of `UserConfigFile` exists and raises a specific error otherwise.

### DIFF
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -65,6 +65,12 @@ func Setup() {
 		}
 	}
 
+	// This may be different from `UserConfigDir` if `--configuration` was provided
+	configDir := filepath.Dir(UserConfigFile)
+	if !utils.Exists(configDir) {
+		utils.HandleError(fmt.Errorf("Configuration file directory does not exist %s", configDir))
+	}
+
 	if !utils.Exists(UserConfigFile) {
 		v1ConfigA := filepath.Join(utils.ConfigDir(), configFileName)
 		v1ConfigB := filepath.Join(utils.HomeDir(), configFileName)


Piccirello:
The commit messages we use in this repo appear directly as changelog entries in the CLI. We should use a commit message that's more use friendly. Something like "Display error when specified configuration dir doesn't exist".

pkg/configuration/config.go
```diff
@@ -65,6 +65,12 @@ func Setup() {
 		}
 	}
 
+	// This may be different from `UserConfigDir` if `--configuration` was provided
+	configDir := filepath.Dir(UserConfigFile)
+	if !utils.Exists(configDir) {
+		utils.HandleError(fmt.Errorf("Parent directory must exist when providing a custom config file path: %s", configDir))
```
Piccirello: I don't want to rely on implicit knowledge that the path must've been overridden, unless we're explicitly checking that a custom path was used
```suggestion
		utils.HandleError(fmt.Errorf("Configuration file directory does not exist %s", configDir))
```

nmanoogian: Good point 👍 I'll make the change as part of my commit amend

pkg/configuration/config.go
```diff
@@ -65,6 +65,12 @@ func Setup() {
 		}
 	}
 
+	// This may be different from `UserConfigDir` if `--configuration` was provided
+	configDir := filepath.Dir(UserConfigFile)
+	if !utils.Exists(configDir) {
+		utils.HandleError(fmt.Errorf("Configuration file directory does not exist: %s", configDir))
```
Piccirello: For consistency with our other errors
```suggestion
		utils.HandleError(fmt.Errorf("Configuration file directory does not exist %s", configDir))
```

nmanoogian: Ah, good catch

Piccirello (approved):

PR #179, opened 2021-03-01
Updated install.sh to work with aarch64 for M1 MacOS

aaronspindler:


Piccirello:
What machine and OS are you currently using? On my M1 MacBook Air running 11.2.1, `uname -m` returns `arm64`.

aaronspindler:
Im on a Mac mini m1, it looks like my uname also returns arm64. However when I run the install script I am getting 
```
"ERROR: Unsupported architecture 'aarch64'  "
```

Piccirello:
Could you run the script again with the `--debug` flag and post the output here? That should give us a bit more info to further investigate.

aaronspindler:
The script is being ran inside of a docker container though, so this could be where the issue is happening.
```
RUN (curl -Ls https://cli.doppler.com/install.sh || wget -qO- https://cli.doppler.com/install.sh) | sh
```

scripts/install.sh
```diff
@@ -96,6 +96,8 @@ elif [ "$uname_machine" = "armv7" ] || [ "$uname_machine" = "armv7l" ]; then
 # armv8?
 elif [ "$uname_machine" = "arm64" ]; then
   arch="arm64"
+elif [ "$uname_machine" = "aarch64" ]; then
+  arch="arm64"
```
Piccirello: Can you combine these two lines and squash into one commit?
```suggestion
elif [ "$uname_machine" = "arm64" ] || [ "$uname_machine" = "aarch64" ]; then
  arch="arm64"
```

Piccirello (commented):
Please prepend your commit message with "chore: "

scripts/install.sh
aaronspindler: I unfortunately can't get my squash working, but I have combined the two lines and added chore: to the beginning of my last commit message.

Piccirello: You can undo your 3 commits with `git reset HEAD~3` and then re-add your changes in one commit. We try really hard to keep a clean commit history in this repo, in part because the version changelogs are generated directly from the commit messages.

aaronspindler: Perfect, I fixed it. Thank you

Piccirello (approved):

Piccirello:
Thanks for this contribution! The fix should now be live.

PR #180, opened 2021-03-03
Add code signing public key to repo

Piccirello:
This provides an alternative method for users to download our public key for signature verification. The primary method is via `gpg --keyserver keyserver.ubuntu.com --recv D3D593D50EE79DEC`.


Piccirello:
Adding a group as a reviewer seems to trigger some very odd behavior...

nmanoogian:
It looks like `doppler_3.23.0_macOS_amd64.pkg` doesn't have a corresponding signature file. Is that because it's already signed for macOS?

Piccirello:
That package is currently built differently than the rest of our binaries. It's going away very soon so you can ignore that.

rgmvisser (approved):

PR #183, opened 2021-03-03
Redact token when printed on screen

Piccirello:
When the token is printed it's generally to let the user know that it has *some* value, rather than needing to expose the exact value. We now display the token in redacted form when it's shown on-screen, with the exception of `doppler configure get token`, `doppler configure debug`, and commands run with the `--json` flag.

Closes DPLR-795.

### DIFF
--- a/pkg/cmd/configure.go
+++ b/pkg/cmd/configure.go
@@ -60,7 +60,7 @@ and your config file. Flags have the highest priority; config file has the least
 		utils.Log(fmt.Sprintf("%s %s", color.Green.Render("Configuration file:"), configuration.UserConfigFile))
 
 		config := configuration.LocalConfig(cmd)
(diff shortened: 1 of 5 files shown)


rgmvisser (dismissed):
Looks good, 1 suggestions: instead of showing redacted, can we show something like:
`dp.pt.BklJ......wdd7iNE8wKD`? That way the token can still be identified if it is a certain token.

nmanoogian (dismissed):

Piccirello:
@rgmvisser good idea, I've modified the redaction logic.

Before:
```
$ ./doppler configure | grep token
│ token          │ [REDACTED]                    │ / │
```

After:
```
$ ./doppler configure | grep token
│ token          │ dp.ct.r126…XmlcH              │ / │
```

rgmvisser (approved):

PR #185, opened 2021-03-07
Auto install shell completions on CLI install

Piccirello:


.goreleaser.yml
```diff
@@ -136,6 +145,11 @@ nfpms:
     formats:
       - deb
       - rpm
+    contents:
+      - src: ./completions/doppler.bash
+        dst: /etc/bash_completion.d/doppler
```
ryan-blunden, line 150: I honestly don't know much about bash-completion, or the install/activation flow, but do we need to take into account env vars such as `$BASH_COMPLETION_DIR`, or if on macOS looking for directory paths such as `/usr/local/etc/bash_completion.d` that don't require sudo/root access to install?

Piccirello: I don't know very much about bash-completion either, I wasn't even aware of its custom environment variables. I'm not sure how we would accomplish that with our current tooling to be honest.

This section is only for the deb and rpm repositories, so it'll be used exclusively on linux. It's worth noting that the binary is installed to `/usr/bin` and thus already requires sudo access.

nmanoogian (approved):

.goreleaser.yml
rgmvisser: Can we take a look at maybe another good open source CLI package that has autocompletion? Maybe we can get some best practices from there?

Piccirello: This change was unabashedly borrowed from goreleaser's own [config](https://github.com/goreleaser/goreleaser/blob/master/.goreleaser.yml). I neglected to document this anywhere other than the linear task (which isn't linked here).

rgmvisser (approved):

PR #187, opened 2021-03-10
Enforce minimum of TLSv1.2 for all requests

Piccirello:


nmanoogian (dismissed):

rgmvisser (dismissed):
Looks good, just wanna make sure this has not potential implications for current users? Would this break if a user does not have TLSv1.2? Maybe we are okay with that?

Piccirello:
Any user with a moderately modern operating system will have access to TLSv1.2. Our domains already enforce a minimum of TLSv1.2, so this shouldn't break anything or anyone.

Piccirello:
Force pushed to solve a merge conflict.

nmanoogian (dismissed):

rgmvisser (dismissed):

scripts/install.sh
```diff
@@ -158,11 +164,17 @@ key_url="https://cli.doppler.com/keys/public"
...
-    echo "Signature verification is an additional measure to ensure you're executing code that Doppler produced"
-    echo "You can remove this warning by installing your system's gnupg package, or by specifying --no-verify-signature"
+    if [ "$FORCE_VERIFY_SIGNATURE" -eq 1 ]; then
+      echo "ERROR: Unable to find gpg binary for signature verficiation"
+      echo "You can resolve this error by installing your system's gnupg package"
+      clean_exit 1
+    else
+      log_debug "Unable to find gpg binary, skipping signature verification"
+      VERIFY_SIGNATURE=0
+      echo "WARNING: Skipping signature verification due to no available gpg binary"
```
rgmvisser, line 174: Just want to make sure that we are okay with printing this. Not sure if this would potentially interfere for user that don't expect any output from our install script (probably not, but just checking).

rgmvisser (approved):
1 questions, but lgtm

scripts/install.sh
Piccirello: I like your thinking. This script already outputs arbitrary text (though we try to minimize how much) so we should be ok to add this.

PR #188, opened 2021-03-13
Add install.sh flag for forcing signature verification

Piccirello:
For users that explicitly want to opt-in to signature verification (like Doppler), we now support a `--verify-signature` flag. This flag will eventually be deprecated when we switch signature verification form opt-in to opt-out.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -75,6 +75,44 @@ jobs:
       run: sudo ./scripts/install.sh --debug
     - name: Test CLI
       run: doppler --version
+  ubuntu-force-verify-signature:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        path: ./src/github.com/${{ github.repository }}
+    - name: Install CLI
+      run: sudo ./scripts/install.sh --debug --verify-signature
+    - name: Test CLI
+      run: doppler --version
+  ubuntu-force-verify-signature-fail:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -90,6 +91,11 @@ for arg; do
     VERIFY_SIGNATURE=0
     echo "Disabling signature verification, this is not recommended"
   fi
+
+  if [ "$arg" = "--verify-signature" ]; then
```
nmanoogian, line 95: Do we have these flags in our docs anywhere? It might be helpful to explain why we have both `--no-verify-signature` and `--verify-signature` flags. Typically, I'd also recommend supporting a `-h`/`--help` flag but I probably wouldn't think to try that on an install one-liner.

.github/workflows/install.yml
```diff
@@ -75,6 +75,44 @@ jobs:
...
+      run: |
+        result=$(which gpg) || true;
+        if [ -n "$result" ]; then
+          exit 1;
+        fi;
+    - name: Install CLI
+      continue-on-error: true
+      run: sudo ./scripts/install.sh --debug --verify-signature
+    - name: Verify install failed
+      if: ${{ always() }}
```
rgmvisser, line 110: Just curious, why is this? :)

rgmvisser (approved):

scripts/install.sh
Piccirello: We traditionally haven't been great about documenting our CLI flags, but it's on our roadmap. Will add to our list of flags to document.

.github/workflows/install.yml
Piccirello: The previous step is expected to fail, so we need this to avoid exiting.

PR #194, opened 2021-04-01
Add Doppler import command

rgmvisser:
This will allow us to read the `doppler.yaml` file and import this into the doppler dashboard with 1 command.

### DIFF
--- a/pkg/cmd/import.go
+++ b/pkg/cmd/import.go
@@ -1,0 +1,69 @@
+/*
+Copyright © 2021 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/printer"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+const projectTemplateFileName = "doppler-template.yaml"
+
+var importCommand = &cobra.Command{
+	Use:   "import",
+	Short: "Import projects into your Doppler workplace",
+	Args:  cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		jsonFlag := utils.OutputJSON
+		localConfig := configuration.LocalConfig(cmd)
+		utils.RequireValue("token", localConfig.Token.Value)
+		projectTemplateFile, err := utils.GetFilePath(cmd.Flag("template").Value.String())
+		if err != nil {
+			utils.HandleError(err, "Unable to parse template file path")
+		}
+		template := readTemplateFile(projectTemplateFile)
+
+		info, importErr := http.ImportTemplate(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, template)
+		if !importErr.IsNil() {
+			utils.HandleError(importErr.Unwrap(), importErr.Message)
+		}
+		printer.ProjectsInfo(info, jsonFlag)
+	},
+}
+
+func readTemplateFile(projectTemplateFile string) []byte {
+	if !utils.Exists(projectTemplateFile) {
+		utils.HandleError(fmt.Errorf("Unable to find project template file: %s", projectTemplateFile))
+	}
+	utils.LogDebug(fmt.Sprintf("Reading template file %s", projectTemplateFile))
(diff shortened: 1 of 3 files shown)


Piccirello:
Does it make sense for doppler.yaml to contain the project template? That seems like a pretty different use case than originally intended for this file.

pkg/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+	Args:  cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		jsonFlag := utils.OutputJSON
+		defaultProjectTemplateFile := filepath.Join("./", projectTemplateFileName)
+		projectTemplateFile := utils.GetPathFlagIfChanged(cmd, "template", defaultProjectTemplateFile)
+
+		template := readTemplateFile(projectTemplateFile)
+
+		localConfig := configuration.LocalConfig(cmd)
+		utils.RequireValue("token", localConfig.Token.Value)
```
Piccirello: Let's check for the token before we read in the project template

pkg/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+	utils.LogDebug(fmt.Sprintf("Reading template file %s", projectTemplateFile))
+	yamlFile, err := ioutil.ReadFile(projectTemplateFile) // #nosec G304
+	if err != nil {
+		utils.HandleError(err, fmt.Sprintf("Unable to read project template file: %s", projectTemplateFile))
+	}
+	return yamlFile
+}
+
+func init() {
+	initCommand.Flags().String("template", "", "template path e.g. './path/to/file.yaml'. (defaults to ./doppler.yaml)")
```
Piccirello: We should specify the default and update the help text

```suggestion
	initCommand.Flags().String("template", filepath.Join("./", projectTemplateFileName), "path to template file (e.g. './path/to/file.yaml')")
```

pkg/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+const projectTemplateFileName = "doppler.yaml"
+
+var initCommand = &cobra.Command{
+	Use:   "init",
+	Short: "Automatically set up projects in your Doppler dashboard",
+	Args:  cobra.MaximumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		jsonFlag := utils.OutputJSON
+		defaultProjectTemplateFile := filepath.Join("./", projectTemplateFileName)
+		projectTemplateFile := utils.GetPathFlagIfChanged(cmd, "template", defaultProjectTemplateFile)
```
Piccirello: Once we're specifying the default value:
```suggestion
		projectTemplateFile, err := utils.GetFilePath(cmd.Flag("template").Value.String())
```

Piccirello (changes requested):

ryan-blunden:
Probably outside the scope of this PR, but [adding URL support for the template param](https://github.com/DopplerHQ/cli/issues/196) would be a killer feature for helping folks use this for setting up open source applications.

pkg/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+	"path/filepath"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/printer"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+// repoConfigFileName (doppler.yaml)
```
Piccirello: I believe this comment is outdated

pkg/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+const projectTemplateFileName = "doppler.yaml"
+
+var initCommand = &cobra.Command{
+	Use:   "init",
+	Short: "Automatically set up projects in your Doppler dashboard",
```
Piccirello: What's the ultimate goal of this command? Is it to support importing a template file, is it to set up a workplace, set up a specific project, etc? I think we can make this description a bit more precise.

ryan-blunden: What about:
 
Scaffold multiple environments and secrets for a repository using a Doppler project template.

rgmvisser: I like that, but you can also setup multiple projects if you would like. What about:
`Scaffold projects, environments and secrets for a repository using a Doppler project template.`

ryan-blunden: Ah right, I wasn't aware you could set up multiple projects.

What's the use case there?

rgmvisser: It's more just built in a scalable way that if this would every be needed in the future this is possible (without having to keep backward compatibility etc).

rgmvisser: 1 use case for instance could be a repo with a server & client. Where you want to have them as 2 projects in Doppler.

rgmvisser: Resolved offline

pkg/cmd/import.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright © 2019 Doppler <support@doppler.com>
```
Piccirello: 2021

Piccirello (approved):

doppler.yaml
```diff
@@ -1,3 +1,27 @@
...
+  - name: doppler-cli
+    description: Official Doppler CLI
+    environments:
+      - slug: dev
+        name: Development
+        configs:
+          - slug: dev
+          - slug: dev_ruud
+      - slug: ci
+        name: Continues Integration
```
Piccirello:
```suggestion
        name: Continuous Integration
```

Piccirello (changes requested):
Let's not use doppler.yaml for this

Piccirello (approved):
LGTM

PR #215, opened 2021-05-05
Throw an error if a non-existent secret is requested from secrets get

apazzolini:
If you request a secret that doesn't exist, we currently return an empty string. This PR changes the behavior to throw an error if the secret was not found. If you request multiple secrets, they must all exist, or an error will be thrown.

```
» doppler secrets get X
Doppler Error: Could not find requested secret: X

» doppler secrets get DOPPLER_PROJECT X
Doppler Error: Could not find requested secret: X

» doppler secrets get X Y
Doppler Error: Could not find requested secrets: X, Y
```

Closes ENG-1753.

### DIFF
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -138,6 +138,7 @@ func getSecrets(cmd *cobra.Command, args []string) {
 	plain := utils.GetBoolFlag(cmd, "plain")
 	copy := utils.GetBoolFlag(cmd, "copy")
 	raw := utils.GetBoolFlag(cmd, "raw")
+	exitOnMissingSecret := !utils.GetBoolFlag(cmd, "no-exit-on-missing-secret")
 	localConfig := configuration.LocalConfig(cmd)
 
 	utils.RequireValue("token", localConfig.Token.Value)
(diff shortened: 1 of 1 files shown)


Piccirello:
This looks great! There's one minor issue: this is a breaking change. If you were previously requesting a non-existent secret, you'd get a blank string, but now you'll get an error. Given the very specific use case I think we may be able to release this as a minor version bump without breaking too many people's flows. Though we may want to add a flag so that if a user's setup does break, they have recourse to restore the original behavior. Wdyt about a new `no-exit-on-missing-secret` flag (or something similarly named)?

apazzolini:
Definitely agree with that change. The flag name is good, matches `no-exit-on-write-failure`.

pkg/printer/print.go
```diff
@@ -249,7 +249,7 @@ func ProjectInfo(info models.ProjectInfo, jsonFlag bool) {
 }
 
 // Secrets print secrets
-func Secrets(secrets map[string]models.ComputedSecret, secretsToPrint []string, jsonFlag bool, plain bool, raw bool, copy bool) {
+func Secrets(secrets map[string]models.ComputedSecret, secretsToPrint []string, jsonFlag bool, plain bool, raw bool, copy bool, exitOnMissingSecret bool) {
```
Piccirello: This `Secrets()` function lives in a package that's primarily intended to print values. It might be more appropriate for the caller to perform this check.

apazzolini: 👍  I should have done that from the get-go.

Piccirello (approved):

nmanoogian (approved):

Issue #210, opened 2021-05-01
Unable to install the latest Doppler CLI [BUG]


joshuachough:
**Describe the bug**
I can't update to the latest Doppler CLI.

**To Reproduce**
Steps to reproduce the behavior. Please include output from running the command with `--debug`.
1. In a console, type `doppler update`
2. After it prints `? Install Doppler CLI v3.24.3 (Y/n)`, type `y`
3. See error (under **Screenshots**)

**Expected behavior**
I expect it to update the Doppler CLI.

**Screenshots**
```
? Install Doppler CLI v3.24.3 Yes
Updating...
Unable to install the latest Doppler CLI
Doppler Error: exit status 2
```

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04.1 LTS

**CLI Version:**
Version 3.23.2

**Additional context**
No idea why this is happening. What does exit status 2 mean?

Piccirello:
Please re-run the command with the `--debug` flag and share the output here.

```sh
doppler update --debug
```

joshuachough:
> Please re-run the command with the `--debug` flag and share the output here.
> 
> ```shell
> doppler update --debug
> ```

When I ran the command with the `--debug` flag it output

```
itsjoshthedeveloper@chough:~/coursetable$ doppler update --debug
Debug: Using config file /home/itsjoshthedeveloper/.doppler/.doppler.yaml
Debug: Reading config file
Debug: Checking for latest version of the CLI
Debug: Performing HTTP GET to https://api.github.com/repos/DopplerHQ/cli/releases/latest
Updating...
Debug: Performing HTTP GET to https://cli.doppler.com/install.sh
Debug: Writing to temp file /home/itsjoshthedeveloper/.install.sh.424238512
Debug: Executing install script
Debug: Executing "/home/itsjoshthedeveloper/.install.sh.424238512 --debug"
DEBUG: Detected OS 'linux'
DEBUG: Detected architecture 'amd64'
DEBUG: Detected format 'deb'
DEBUG: Checking for gpg binary
DEBUG: Using temp directory /home/itsjoshthedeveloper/.tmp.vqHXYQZe
Downloading Doppler CLI
DEBUG: Using /usr/bin/curl for requests
DEBUG: Downloading binary from https://cli.doppler.com/download?os=linux&arch=amd64&format=deb
DEBUG: Downloaded CLI v3.24.3
[...]

PR #218, opened 2021-05-23
Print warning when doppler update fails due to lack of perms

Piccirello:
Addresses #210, closes ENG-2206

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -65,7 +65,13 @@ func RunInstallScript() (bool, string, Error) {
 		fmt.Println(strOut)
 	}
 	if err != nil {
-		return false, "", Error{Err: err, Message: "Unable to install the latest Doppler CLI"}
+		message := "Unable to install the latest Doppler CLI"
+		// check for errors indicating lack of perms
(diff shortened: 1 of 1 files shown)


pkg/controllers/update.go
```diff
@@ -65,7 +65,13 @@ func RunInstallScript() (bool, string, Error) {
 		fmt.Println(strOut)
 	}
 	if err != nil {
-		return false, "", Error{Err: err, Message: "Unable to install the latest Doppler CLI"}
+		message := "Unable to install the latest Doppler CLI"
+		// check for errors indicating lack of perms
+		if strings.Contains(strOut, "dpkg: error: requested operation requires superuser privilege") {
```
nmanoogian, line 70: When this error occurs, does dpkg return an EACCES (13) error code? It might be a bit cleaner to check that instead, if possible. Otherwise, I think the error message check is better than nothing 👍

Piccirello: That's a good point. The tricky thing is that that error would be captured inside the `install.sh` script, but that error code isn't bubbled up. I think for now this might be the easiest solution.

nmanoogian (approved):

PR #219, opened 2021-05-25
Add support for templating with `doppler secrets template`

nmanoogian:


pkg/cmd/secrets.go
```diff
@@ -380,6 +408,89 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+
+	var outputFile *os.File
+	var outputFilePath string
+	output := cmd.Flag("output").Value.String()
+	if len(output) != 0 {
+		outputFilePath, err = utils.GetFilePath(output)
+		if err != nil {
+			utils.HandleError(err, "Unable to parse output file path")
+		}
+		outputFile, err = os.Create(outputFilePath)
```
Piccirello: Can we avoid creating the file until the secrets fetch has succeeded? In the event that secrets fetch fails, we'll have created a blank file.

pkg/cmd/secrets.go
```diff
@@ -380,6 +408,89 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+
+	secretsMap := map[string]interface{}{}
+	for _, secret := range secrets {
+		secretsMap[secret.Name] = secret.ComputedValue
+	}
+
+	if outputFile != nil {
+		err = template.Execute(outputFile, secretsMap)
+		if err != nil {
+			utils.HandleError(parseErr, "Unable to render template")
```
Piccirello: When testing locally using the provided template, my config didn't have the correct secrets. This was the output, which is a little confusing:

```
$ ./doppler secrets template template.yaml
host: <no value>
port: <no value>



Multiline: Unable to render template
Doppler Error: <nil>
```

pkg/cmd/secrets.go
```diff
@@ -380,6 +408,89 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+		utils.HandleError(parseErr, "Unable to parse API response")
+	}
+
+	secretsMap := map[string]interface{}{}
+	for _, secret := range secrets {
+		secretsMap[secret.Name] = secret.ComputedValue
+	}
+
+	if outputFile != nil {
+		err = template.Execute(outputFile, secretsMap)
```
Piccirello: I think this command should be atomic- we either write the completed template, or we write nothing. It seems this is counter to how `Execute` works though: `If an error occurs executing the template or writing its output, execution stops, but partial results may already have been written to the output writer.`

Do you have any ideas how we can get around this? I'm thinking maybe we write to a local writer and then if that succeeds we write the contents to the specified file/stdout. This would also solve for the scenario I mentioned above, where the stdout output was confusing.

Piccirello (commented):
This looks really cool, and I think it solves ENG-1114. I just want to tweak the error handling a little bit.

pkg/cmd/secrets.go
nmanoogian: Great call 👍

pkg/cmd/secrets.go
nmanoogian: Yeah, I noticed this too. There is an option that I can add with `.Option("missingkey=error")` but if we do that, we won't be able to do `{{ if .MISSING_KEY }}` which I think could actually be really valuable.

nmanoogian: We could add a flag to enable/disable that. What do you think?

nmanoogian: Oh wait, I missed the partial write issue on my first read. Let me see what can be done about that

pkg/cmd/secrets.go
nmanoogian: Yeah, this is a little weird for sure. It seems like it might be done this way to allow the templating to work on very large files that wouldn't otherwise fit into memory. I'd prefer to optimize for user experience on smaller files and follow your local-writer suggestion.

pkg/cmd/secrets.go
```diff
@@ -109,6 +111,32 @@ $ doppler secrets download --format=env --no-file`,
...
+var secretsTemplateCmd = &cobra.Command{
+	Use:   "template <filepath>",
+	Short: "Inject secrets into a template file",
+	Long:  "Inject secrets into a template file. See https://golang.org/pkg/text/template/ for full syntax",
+	Example: `$ cat template.yaml
+{{- /* Full comment support */ -}}
+host: {{.API_HOST}}
+port: {{.API_PORT}}
+
+{{with .OPTIONAL_SECRET}}
```
nmanoogian: I wanted to call out this `with` syntax, which is awesome! It creates a new scope with the provided value so you can just refer to the secret as `.` inside the block.

nmanoogian:
@Piccirello Can I get your take on these Salus advisories as well?

pkg/cmd/secrets.go
Piccirello: One nit: I really liked the `if` example! As someone seeing this syntax for the first time, it made intuitive sense how I might use that. Could you add that back as an example?

pkg/cmd/secrets.go
```diff
@@ -380,6 +408,85 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
...
+	}
+
+	buffer := new(strings.Builder)
+	err = template.Execute(buffer, secretsMap)
+	if err != nil {
+		utils.HandleError(err, "Unable to render template")
+	}
+
+	if outputFilePath != "" {
+		err = ioutil.WriteFile(outputFilePath, []byte(buffer.String()), 0644)
```
Piccirello:
```suggestion
		err = ioutil.WriteFile(outputFilePath, []byte(buffer.String()), 0600)
```

This would allow all users to read these secrets 😢

Piccirello (commented):
> @Piccirello Can I get your take on these Salus advisories as well?

It looks like there are 3 gosec errors. The output is a little tough to grok at first but it's mostly just pointing out required code changes. The most relevant fields are `details` and `file`. The first issue that's reported `G304 - Potential file inclusion via variable` is likely intentional behavior. You can see how we ignore these by searching the codebase for `nosec`. The other two should be fixed.

pkg/cmd/secrets.go
Piccirello: Tested this new functionality and it works exactly as expected. Nice job!

pkg/cmd/secrets.go
nmanoogian: Oh dear -- Egg on my face, sorry about that

pkg/cmd/secrets.go
nmanoogian: That's a good point. People who really want to power-use this feature can dig into the Go docs and learn about `with`.

nmanoogian:
> It looks like there are 3 gosec errors. The output is a little tough to grok at first but it's mostly just pointing out required code changes. The most relevant fields are `details` and `file`. The first issue that's reported `G304 - Potential file inclusion via variable` is likely intentional behavior. You can see how we ignore these by searching the codebase for `nosec`. The other two should be fixed.

Ah, G304 was the one I was most confused about. The other two are straightforward but I had a different batch of advisories before I made the buffer changes. I'll make the necessary updates now 👍

nmanoogian:
@Piccirello Do you think it would be worth it to add a `--delimiter` option to let people pick between `{{ }}`, `{ }`, `${ }`, `<% %>`, etc.?

pkg/cmd/secrets.go
Piccirello: All good, that's why we have gosec :)

Piccirello:
> @Piccirello Do you think it would be worth it to add a `--delimiter` option to let people pick between `{{ }}`, `{ }`, `${ }`, `<% %>`, etc.?

Hmm, good question. I think we can add that later since it would be backwards compatible. Let's get a couple users first and have someone request this.

Piccirello (dismissed):

Piccirello:
One thought on naming: what do you think of `doppler secrets substitute` (or some other verb). My concern with `template` is that we have project templates, which sound similar but are completely different.

nmanoogian:
> Hmm, good question. I think we can add that later since it would be backwards compatible. Let's get a couple users first and have someone request this.

Yep, I was thinking the same 👍

> One thought on naming: what do you think of doppler secrets substitute (or some other verb). My concern with template is that we have project templates, which sound similar but are completely different.

Oh, that's a great point. Here are some options off the top of my head:

- substitute / sub
- inject
- replace
- format (inspired by python's format function but maybe sounds too much like indentation, etc)
- interpolate / interp

I'm kinda leaning towards substitute or inject. My main hesitation with substitute is that you can do quite a bit more with this that just a plain 'ole find-and-replace.

rgmvisser:
> substitute / sub
inject
replace
format (inspired by python's format function but maybe sounds too much like indentation, etc)
interpolate / interp

My vote would go to substitute or inject 👍

Piccirello (approved):
🎉

PR #224, opened 2021-06-09
Fix double update check when running 'doppler update'

Piccirello:
Closes ENG-2221

### DIFF
--- a/pkg/cmd/enclave_secrets.go
+++ b/pkg/cmd/enclave_secrets.go
@@ -120,6 +120,14 @@ func init() {
 	enclaveSecretsDownloadCmd.Flags().String("format", models.JSON.String(), "output format. one of [json, env]")
 	enclaveSecretsDownloadCmd.Flags().String("passphrase", "", "passphrase to use for encrypting the secrets file. the default passphrase is computed using your current configuration.")
 	enclaveSecretsDownloadCmd.Flags().Bool("no-file", false, "print the response to stdout")
+	// fallback flags
+	enclaveSecretsDownloadCmd.Flags().String("fallback", "", "path to the fallback file. encrypted secrets are written to this file after each successful fetch. secrets will be read from this file if subsequent connections are unsuccessful.")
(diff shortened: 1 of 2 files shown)


pkg/cmd/enclave_secrets.go
```diff
@@ -120,6 +120,14 @@ func init() {
 	enclaveSecretsDownloadCmd.Flags().String("format", models.JSON.String(), "output format. one of [json, env]")
 	enclaveSecretsDownloadCmd.Flags().String("passphrase", "", "passphrase to use for encrypting the secrets file. the default passphrase is computed using your current configuration.")
 	enclaveSecretsDownloadCmd.Flags().Bool("no-file", false, "print the response to stdout")
+	// fallback flags
+	enclaveSecretsDownloadCmd.Flags().String("fallback", "", "path to the fallback file. encrypted secrets are written to this file after each successful fetch. secrets will be read from this file if subsequent connections are unsuccessful.")
```
Piccirello, line 124: The `enclave secrets download` command currently panics when called due to these flags not being defined. It appears this has been broken since [v3.15.0](https://github.com/DopplerHQ/cli/releases/tag/3.15.0), which is when we introduced fallback support to `secrets download`. Given that we haven't heard from any customers, I would surmise that our deprecation message was successful in getting people to switch to the new command.

nmanoogian (approved):

pkg/cmd/root.go
```diff
@@ -69,9 +69,14 @@ var rootCmd = &cobra.Command{
 }
 
 func checkVersion(command string) {
-	// disable version checking on the "run" command and "secrets download" command
-	if command == "run" || command == "download" {
-		return
+	// disable version checking on commands commonly used in production workflows
+	// also disable when explicitly calling 'update' command to avoid checking twice
+	disabledCommands := []string{"run", "secrets download", "update"}
```
rgmvisser, line 74: Check to update when calling `doppler update` 😂

rgmvisser (approved):

PR #229, opened 2021-06-18
Improve error message when token revocation fails

Piccirello:
Closes ENG-2483


rgmvisser (approved):

nmanoogian (approved):

Piccirello:
Will ship this once the corresponding server-side change is live, otherwise we'll end up showing people an unhelpful error message in the mean time.

PR #233, opened 2021-06-25
Add support for auto-completing command args

Piccirello:
Closes ENG-434, closes ENG-2578

To test:
1. Build branch or download [precompiled binary](https://github.com/DopplerHQ/cli/suites/3093700314/artifacts/70679331)
2. Install latest shell completions with `./doppler completion install` (this step will be performed automatically when users run `doppler update`)
3. Open a new shell
4. Try it out: type `doppler secrets get ` and hit `[tab][tab]`


pkg/cmd/root.go
```diff
@@ -113,6 +117,11 @@ func checkVersion(command string) {
 	configuration.SetVersionCheck(versionCheck)
 }
 
+// persistentValidArgsFunction Cobra parses flags after executing ValidArgsFunction, so we must manually initialize flags
+func persistentValidArgsFunction(cmd *cobra.Command) {
+	loadFlags(cmd)
+}
```
Piccirello, line 123: More info: https://github.com/spf13/cobra/issues/1291

nmanoogian:
I'm getting an error on zsh from a local build:
```
./doppler completion install
Doppler Error: Your shell is not supported
```

pkg/cmd/activity.go
```diff
@@ -44,9 +45,10 @@ var activityCmd = &cobra.Command{
 }
 
 var activityGetCmd = &cobra.Command{
-	Use:   "get [log_id]",
-	Short: "Get workplace activity log",
-	Args:  cobra.MaximumNArgs(1),
+	Use:               "get [log_id]",
+	Short:             "Get workplace activity log",
+	Args:              cobra.MaximumNArgs(1),
+	ValidArgsFunction: activityLogIDsValidArgs,
```
nmanoogian, line 51: Does `ValidArgsFunction` do validation as well? I'm mainly concerned about a situation where `controllers.GetActivityLogsIDs` returns the first N results and we're trying to provide an ID for log N+1.

Piccirello:
> I'm getting an error on zsh from a local build:

Can you build the current `master` branch and see if you also get that error? I'm wondering if that's new or pre-existing.

pkg/cmd/activity.go
Piccirello: `ValidArgsFunction` is purely additive, so it will only provide auto-completion when it can. It won't prevent the user from using an arg it doesn't know about.

nmanoogian: Awesome 👍

nmanoogian:
> Can you build the current `master` branch and see if you also get that error? I'm wondering if that's new or pre-existing.

Ah, it looks like it's doing it on master too.

pkg/cmd/configure.go
```diff
@@ -119,6 +120,7 @@ var configureSetCmd = &cobra.Command{
 
 Ex: set the options "key" and "otherkey":
 doppler configure set key=123 otherkey=456`,
+	ValidArgsFunction: configOptionsValidArgs,
```
rgmvisser, line 123: Should this also be `currentConfigOptionsValidArgs`?

pkg/cmd/configure.go
```diff
@@ -227,6 +230,36 @@ var configureResetCmd = &cobra.Command{
 	},
 }
 
+// currentConfigOptionsValidArgs the options currently in use in the current scope
```
rgmvisser, line 233: This comment seems to be for the `currentConfigOptionsValidArgs` function?

pkg/cmd/root.go
rgmvisser: Can add that link in the code as well? Will make it easier to find back probably

rgmvisser (approved):
Looks good to me, just a few small things. 
I guess if the issue around zsh already exists on master we can fix that in a separate PR.

pkg/cmd/configure.go
Piccirello: We want to list all possible config options here, rather than only the options that are already set.

pkg/cmd/configure.go
Piccirello: Will update it

PR #234, opened 2021-06-28
Add `open docs` command

ryan-blunden:
Make it easier to open our docs without manually opening in the browser.

Closes ENG-2588

### DIFF
--- a/pkg/cmd/open.go
+++ b/pkg/cmd/open.go
@@ -80,10 +80,23 @@ var openGithubCmd = &cobra.Command{
 	},
 }
 
+var openDocsCmd = &cobra.Command{
+	Use:   "docs",
+	Short: "open the Doppler documentation home page",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		err := open.Run("https://docs.doppler.com/")
+		if err != nil {
+			utils.HandleError(err)
+		}
+	},
+}
+
 func init() {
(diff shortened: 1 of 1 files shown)


pkg/cmd/open.go
```diff
@@ -80,10 +80,23 @@ var openGithubCmd = &cobra.Command{
 	},
 }
 
+var openDocsCmd = &cobra.Command{
+	Use:   "docs",
+	Short: "Open the Doppler documentation home page",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		err := open.Run("https://dashboard.doppler.com/workplace/docs")
```
nmanoogian: For good measure, should this be based on the `DashboardHost`?

rgmvisser: Should we just do `https://docs.doppler.com/`? In case someone is not logged in?

nmanoogian (dismissed):
Nice! Can we squash those commits?

pkg/cmd/open.go
```diff
@@ -80,10 +80,23 @@ var openGithubCmd = &cobra.Command{
 	},
 }
 
+var openDocsCmd = &cobra.Command{
+	Use:   "docs",
+	Short: "Open the Doppler documentation home page",
```
Piccirello: super nit: can you match the capitalization of the other `open` commands

Piccirello (approved):

PR #241, opened 2021-06-29
Automatically size table based on Terminal width

Piccirello:
This change adapts the CLI's output to the current terminal size. It results in much better looking output at smaller sizes.

Closes ENG-2602


pkg/printer/print.go
```diff
@@ -64,10 +75,14 @@ func Table(headers []string, rows [][]string, options tableOptions) {
 	}
 	t.AppendHeader(tableHeaders)
 
+	numCols := numColumns(rows)
+	maxColWidths := maxColWidths(rows, numCols)
+	colWidths := optimalColWidths(maxColWidths, numCols)
```
Piccirello, line 76: These functions might be worth unit testing

pkg/printer/enclave.go
```diff
@@ -0,0 +1,363 @@
+/*
```
Piccirello, line 1: The commit `chore: move enclave print functions to separate file` is purely moving functions- there aren't any logic changes.

nmanoogian (approved):
This is awesome!

rgmvisser (approved):

rgmvisser:
Nice papercut!

apazzolini (approved):
Works well in tmux, nice one!

Piccirello:
> Works well in tmux, nice one!

Thanks for testing that use case!

PR #248, opened 2021-08-06
Use cli.doppler.com to check version

nmanoogian:


nmanoogian:
Small thing -- just fixed the error message that we throw if we can't read the version

pkg/http/github.go
```diff
@@ -40,11 +40,11 @@ func getLatestVersion() (string, error) {
 		return "", err
 	}
 
-	if version, exists := body["tag_name"]; exists {
+	if version, exists := body["version"]; exists {
 		return version.(string), nil
 	}
 
-	return "", errors.New("unable to retrieve tag_name of latest release")
+	return "", errors.New("unable to retrieve 'version' of latest release")
```
Piccirello: This is probably a good opportunity to make this message a little more user friendly

```suggestion
	return "", errors.New("unable to retrieve latest CLI version")
```

Piccirello (dismissed):

pkg/http/github.go
nmanoogian: I'm game!

Piccirello (approved):

PR #255, opened 2021-09-20
[ENG-2523] Add TOB Semgrep rules

zbuc:


pkg/cmd/login.go
```diff
@@ -76,9 +76,18 @@ var loginCmd = &cobra.Command{
 		if !err.IsNil() {
 			utils.HandleError(err.Unwrap(), err.Message)
 		}
-		code := response["code"].(string)
-		pollingCode := response["polling_code"].(string)
-		authURL := response["auth_url"].(string)
+		code, ok := response["code"].(string)
+		if !ok {
+			utils.HandleError(fmt.Errorf("Unexpected type mismatch for auth code, expected string, got %T", response["code"]))
+		}
```
Piccirello: We probably don't want to display these errors to the user since they're not actionable. We can instead display a generic error ("Unable to parse API response") and then print more helpful information as a debug message.
```suggestion
		code, ok := response["code"].(string)
		if !ok {
		        utils.LogDebug(fmt.Sprintf("Unexpected type mismatch for auth code, expected string, got %T", response["code"]))
			utils.HandleError(errors.New("Unable to parse API response"))
		}
```

zbuc: OK I think this should be better now.

Piccirello (approved):

PR #258, opened 2021-09-23
Add support for name transformers

nmanoogian:
Closes ENG-3194

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -102,7 +102,16 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 			}
 		}
 
-		secrets := fetchSecrets(localConfig, enableCache, enableFallback, fallbackPath, legacyFallbackPath, metadataPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase)
+		nameTransformerString := cmd.Flag("name-transformer").Value.String()
+		var nameTransformer *models.SecretsNameTransformer
+		if nameTransformerString != "" {
+			nameTransformer = models.SecretsNameTransformerMap[nameTransformerString]
+			if nameTransformer == nil || !nameTransformer.EnvCompat {
+				utils.HandleError(fmt.Errorf("invalid name transformer. Valid transformers are %s", validEnvCompatNameTransformersList))
+			}
+		}
+
+		secrets := fetchSecrets(localConfig, enableCache, enableFallback, fallbackPath, legacyFallbackPath, metadataPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase, nameTransformer)
 
 		if preserveEnv {
 			utils.LogWarning("Ignoring Doppler secrets already defined in the environment due to --preserve-env flag")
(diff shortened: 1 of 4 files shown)


pkg/models/secrets_name_transformer.go
```diff
@@ -0,0 +1,58 @@
...
+	Type:      "dotnet",
+	EnvCompat: false,
+}
+var DotNETENVTransformer = &SecretsNameTransformer{
+	Name:      ".NET (ENV)",
+	Type:      "dotnet-env",
+	EnvCompat: true,
+}
+
+var SecretsNameTransformersList = []*SecretsNameTransformer{
```
rgmvisser, line 55: Super nit: why not put this in a map for easy lookup?

pkg/cmd/run.go
```diff
@@ -293,7 +317,7 @@ func fetchSecrets(localConfig models.ScopedOptions, enableCache bool, enableFall
 		utils.HandleError(err, "Unable to parse API response")
 	}
 
-	writeFallbackFile := enableFallback && !fallbackReadonly
+	writeFallbackFile := enableFallback && !fallbackReadonly && nameTransformer == nil
```
rgmvisser, line 308: We don't support fallback files with the name transformer?

rgmvisser (dismissed):
Looks good from my side, just a couple of small Qs.

pkg/models/secrets_name_transformer.go
nmanoogian: I like that better 👍

pkg/cmd/run.go
nmanoogian: Mainly because it's technically challenging and we want to gauge interest on this feature before we fully build it out.

Specifically, if you run `doppler run -- env`, it'll save the resulting secrets to the fallback file without any name transformation. If you then run `doppler run --name-transform camel --fallback-only -- env`, it will provide you with the un-transformed secrets without an error. In order to fail properly in that case, we'd need to provide some info in the fallback file (or the path itself) which include the name transformer which was initially used. I initially planned to add it to the metadata file but supplying `--fallback-only` will bypass the cache check today.

pkg/models/secrets_name_transformer.go
```diff
@@ -0,0 +1,61 @@
...
+var CamelTransformer = &SecretsNameTransformer{
+	Name:      "Camel",
+	Type:      "camel",
+	EnvCompat: true,
+}
+var DotNETTransformer = &SecretsNameTransformer{
+	Name:      ".NET",
+	Type:      "dotnet",
+	EnvCompat: false,
+}
```
Piccirello, line 48: This transformer will produce keys that violate the env var spec. Should we exclude it from `run`?

Piccirello (commented):
Are we able to have customers start testing this before we ship the PR (mostly a question for @ryan-blunden)? We'll also want to capture somewhere that this is a beta feature so that we can remove it without having to perform a major version bump.

pkg/models/secrets_name_transformer.go
nmanoogian: Absolutely 👍 The `run` command handler throws an error if you try to select a name transformer which has `EnvCompat` set to false. I just pushed another change which makes this clearer from the `--help` text.

nmanoogian:
Force push to rebase on develop and fix conflicts

nmanoogian:
Force push adds "BETA" messaging to the `--name-transformer` flag

Piccirello (dismissed):

nmanoogian:
Force push

- Allows `--name-transformer` or `-n` shorthand
- Adds `lower-snake` name transformer to list

nmanoogian:
Force push removed unnecessary whitespace change

pkg/cmd/secrets.go
```diff
@@ -620,6 +631,7 @@ func init() {
 	secretsDownloadCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
 	secretsDownloadCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	secretsDownloadCmd.Flags().String("format", models.JSON.String(), fmt.Sprintf("output format. one of %s", validFormatList))
+	secretsDownloadCmd.Flags().StringP("name-transformer", "n", "", fmt.Sprintf("(BETA) output name transformer. one of %v", validNameTransformersList))
```
Piccirello: We should hold off on adding the `-n` short flag since we may want to use that for something else (espec. something global).

nmanoogian: Could we pick a different short flag? I'm a little worried that `--name-transformer` is just way too long to be useful.

Some alternatives:

- `-x` for a "cross" mnemonic
- `-r` for a "rename" mnemonic

nmanoogian: Another alternative could be to drop the short flag and make the "long flag" something shorter like `--rename`

Piccirello: Do we expect users to be running this ad hoc or will this mostly be in scripts?

nmanoogian: I'm not sure yet but the biggest use case I'm thinking of right now is Terraform:

```
doppler run --name-transformer tf-var -- terraform apply
# vs
doppler run -x tf-var -- terraform apply
```

Personally, if I were able to do that second command, I'd probably just do it ad hoc every time. If I had to type `--name-transformer`, I'd probably write an alias or dig it out with `ctrl-r` whenever I need it

Piccirello: A short flag can always be added later, but once added it can't be removed. I think we should ship without one for now and get some usage before adding it.

nmanoogian: Copy that! I'll remove the short-flag for now

nmanoogian:
Force push removes short flag for `--name-transformer`

Piccirello (approved):

PR #260, opened 2021-09-28
Add doppler secrets edit command

apazzolini:
This works pretty well from my testing, but there are more edge cases to handle before this is ready to merge. At a minimum, this needs:

- [ ] Tests around the parsing logic
- [ ] Tests around edge cases (like saving an empty file or a file that we can't parse)
- [ ] Delete secret support? (compare against the ones you chose, delete any not present in the edited version)
- [ ] Better copy and help text
- [ ] Research if Windows has a way to edit a file that's in memory and not on disk

Demo workflow:

https://user-images.githubusercontent.com/869471/135178810-502ed8c8-8a85-47b3-b243-a5db19d13e96.mov

### DIFF
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -139,6 +139,15 @@ JSON Secret: "{\"logging\": \"info\"}"`,
 	Run:  substituteSecrets,
 }
 
+var secretsEditCmd = &cobra.Command{
+	Use:     "edit",
+	Short:   "Edit secrets in your editor",
+	Long:    "Shows a list of current secrets in the config and allows you to edit in your $EDITOR",
+	Example: `$ doppler secrets edit`,
+	Args:    cobra.NoArgs,
+	Run:     editSecrets,
+}
+
 func secrets(cmd *cobra.Command, args []string) {
 	jsonFlag := utils.OutputJSON
 	raw := utils.GetBoolFlag(cmd, "raw")
@@ -552,6 +561,8 @@ func init() {
 	secretsSubstituteCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	secretsSubstituteCmd.Flags().String("output", "", "path to the output file. by default the rendered text will be written to stdout.")
 	secretsCmd.AddCommand(secretsSubstituteCmd)
+
+	secretsCmd.AddCommand(secretsEditCmd)
 
 	rootCmd.AddCommand(secretsCmd)
 }
--- a/pkg/cmd/secrets_edit_linux_amd64.go
+++ b/pkg/cmd/secrets_edit_linux_amd64.go
@@ -1,0 +1,219 @@
+/*
+Copyright © 2020 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"bufio"
+	"fmt"
+	"golang.org/x/sys/unix"
+	"os"
+	"os/exec"
+	"sort"
+	"strings"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/models"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var header = strings.TrimPrefix(`
+┌───────────────────────────────────────────────────────────────────────┐
+│ Saving this file and exiting will apply your changes.                 │
+│ Exiting without saving will cancel the operation.                     │
+│                                                                       │
+│ If a secret is removed from this file, it will be removed in Doppler. │
+│                                                                       │
+│ Secrets in this file follow the format:                               │
+│                                                                       │
+│ SECRET_ONE                                                            │
+│ the secret value, which can span                                      │
+│ multiple lines, and is terminated by a                                │
+│ blank line followed by a period.                                      │
+│                                                                       │
+│ .                                                                     │
+│ SECRET_TWO                                                            │
+│ the second secret value                                               │
+│                                                                       │
+│ .                                                                     │
+└───────────────────────────────────────────────────────────────────────┘
+
+`, "\n")
+
+func editSecrets(cmd *cobra.Command, args []string) {
+	localConfig := configuration.LocalConfig(cmd)
+	utils.RequireValue("token", localConfig.Token.Value)
+
+	secrets := fetchSecretsForEdit(localConfig)
+	selectedSecrets := selectSecrets(secrets)
+	buffer := mapToEditFormat(selectedSecrets, secrets)
+	modifiedSecrets := EditInEditor(buffer)
+
+	if buffer == modifiedSecrets {
+		utils.HandleError(fmt.Errorf("File was unchanged, not writing secrets."))
+	}
+
+	secretsToSave := parseEditFormat(selectedSecrets, modifiedSecrets)
+	saveSecrets(localConfig, secretsToSave)
+
+	fmt.Println("Secrets have been updated.")
+}
+
+func fetchSecretsForEdit(localConfig models.ScopedOptions) map[string]models.ComputedSecret {
+	response, httpErr := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+	if !httpErr.IsNil() {
+		utils.HandleError(httpErr.Unwrap(), httpErr.Message)
+	}
+
+	secrets, err := models.ParseSecrets(response)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/cmd/secrets.go
```diff
@@ -553,5 +562,7 @@ func init() {
 	secretsSubstituteCmd.Flags().String("output", "", "path to the output file. by default the rendered text will be written to stdout.")
 	secretsCmd.AddCommand(secretsSubstituteCmd)
 
+	secretsCmd.AddCommand(secretsEditCmd)
```
Piccirello, line 565: We'll want to add the `project` and `config` flags for consistency with our other commands.

pkg/cmd/secrets.go
```diff
@@ -139,6 +139,15 @@ JSON Secret: "{\"logging\": \"info\"}"`,
...
 
+var secretsEditCmd = &cobra.Command{
+	Use:     "edit",
+	Short:   "Edit secrets in your editor",
+	Long:    "Shows a list of current secrets in the config and allows you to edit in your $EDITOR",
+	Example: `$ doppler secrets edit`,
+	Args:    cobra.NoArgs,
+	Run:     editSecrets,
+}
+
```
Piccirello, line 150: It would be really cool if this showed a summary of the changes after exiting the viewer, like what secrets were modified, and allowed you to either confirm, cancel, or continue editing. This could also be a future enhancement.

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+	"strings"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/models"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var header = `# Save this file and quit your editor to save all secrets below.
```
Piccirello: nit: this can be const

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+	modifiedSecrets := EditInEditor(buffer)
+
+	if buffer == modifiedSecrets {
+		utils.HandleError(fmt.Errorf("File was unchanged, not writing secrets."))
+	}
+
+	secretsToSave := parseEditFormat(modifiedSecrets)
+	saveSecrets(localConfig, secretsToSave)
+
+	fmt.Print("Secrets have been updated.\n")
```
Piccirello: nit: can we use `fmt.Println` and remove the newline character?

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+			value = strings.TrimSuffix(secretArr[1], "\n")
+		}
+		secretsMap[name] = value
+	}
+
+	return secretsMap
+}
+
+func isEditableSecret(name string) bool {
+	if name == "DOPPLER_CONFIG" || name == "DOPPLER_ENVIRONMENT" || name == "DOPPLER_PROJECT" {
```
Piccirello: We have a couple other legacy managed secrets we'll want to include: `["DOPPLER_ENCLAVE_PROJECT", "DOPPLER_ENCLAVE_ENVIRONMENT", "DOPPLER_ENCLAVE_CONFIG"]`

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+		utils.HandleError(httpErr.Unwrap(), httpErr.Message)
+	}
+}
+
+func mapToEditFormat(selectedSecrets []string, secrets map[string]models.ComputedSecret) string {
+	buffer := header
+
+	for idx, val := range selectedSecrets {
+		buffer = buffer + val
+		buffer = buffer + "\n" + secrets[val].RawValue
```
Piccirello: nit: it's more idiomatic to use `fmt.Sprintf` here

```go
buffer = fmt.Sprintf("%s%s\n%s", buffer, val, ecrets[val].RawValue)
```

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+		utils.HandleError(err, "Unable to launch editor")
+	}
+
+	err = cmd.Wait()
+	if err != nil {
+		utils.HandleError(err, "Unable to wait for editor")
+	}
+
+	content, _ := os.ReadFile(tmpFilePath)
+	defer unix.Close(tmpFd)
```
Piccirello: Since this call is deferred, can we move it to immediately after the fd is returned by `utils.Memfile`? That way an intermediate error doesn't prevent us from closing it.

pkg/utils/io.go
```diff
@@ -76,3 +77,29 @@ func WriteTempFile(name string, data []byte, perm os.FileMode) (string, error) {
 
 	return tmpFileName, nil
 }
+
+func Memfile(name string, b []byte) (int, error) {
+	fd, err := unix.MemfdCreate(name, 0)
+	if err != nil {
+		return 0, fmt.Errorf("MemfdCreate: %v", err)
+	}
```
Piccirello: Is it guaranteed that if this function returns an error it hasn't created the file descriptor? Otherwise we should close it in the error handler.

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+	env := os.Environ()
+	editor := "vim"
+	for _, envVar := range env {
+		// key=value format
+		parts := strings.SplitN(envVar, "=", 2)
+		key := parts[0]
+		if strings.ToUpper(key) == "EDITOR" {
+			editor = parts[1]
+		}
+	}
```
Piccirello: Since this surreptitiously relies on the environment, we may want to introduce out somewhat-standard `no-read-env` flag here.

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+	tmpFd, err := utils.Memfile("fileToEdit", []byte(contents))
+	if err != nil {
+		utils.HandleError(err, "Unable to create file descriptor")
+	}
+
+	// filepath to our newly created in-memory file descriptor
+	tmpFilePath := fmt.Sprintf("/proc/self/fd/%d", tmpFd)
+
+	env := os.Environ()
+	editor := "vim"
```
Piccirello: Is there any risk of vim writing a backup file, swap file, or some other file to disc? I do see your `noswapfile` flag below.

We'll probably want to document the risk of using other text editors, maybe in the description of the command.

pkg/cmd/secrets_edit.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func EditInEditor(contents string) string {
+	tmpFd, err := utils.Memfile("fileToEdit", []byte(contents))
+	if err != nil {
+		utils.HandleError(err, "Unable to create file descriptor")
+	}
+
+	// filepath to our newly created in-memory file descriptor
+	tmpFilePath := fmt.Sprintf("/proc/self/fd/%d", tmpFd)
```
Piccirello: What permissions does this fd exist with? Is there any risk of other processes accessing this file? I'm wondering if we can lock it down more, otherwise this may be some fd that any other process (or at least any other process running as this user) can access. In which case it's no different than just writing it to the filesystem.

pkg/utils/io.go
```diff
@@ -76,3 +77,29 @@ func WriteTempFile(name string, data []byte, perm os.FileMode) (string, error) {
...
+	if err != nil {
+		return 0, fmt.Errorf("Mmap: %v", err)
+	}
+
+	copy(data, b)
+
+	err = unix.Munmap(data)
+	if err != nil {
+		return 0, fmt.Errorf("Munmap: %v", err)
+	}
```
Piccirello: I don't fully understand this pattern, in part because Go doesn't provide any docs on this command, and linux's [mmap](https://www.man7.org/linux/man-pages/man2/mmap.2.html) doesn't take an fd. How is that we can copy to `data` and then immediately unmap it from memory?

pkg/cmd/secrets.go
```diff
@@ -139,6 +139,15 @@ JSON Secret: "{\"logging\": \"info\"}"`,
...
 }
 
+var secretsEditCmd = &cobra.Command{
+	Use:     "edit",
+	Short:   "Edit secrets in your editor",
+	Long:    "Shows a list of current secrets in the config and allows you to edit in your $EDITOR",
+	Example: `$ doppler secrets edit`,
+	Args:    cobra.NoArgs,
+	Run:     editSecrets,
+}
```
Piccirello, line 149: Does this gracefully handle secret renames? If not we should just add a note about that limitation to the help text.

Piccirello (commented):
This is really f'in cool

pkg/utils/io.go
apazzolini: Great question. The snippet came from [this blog post](https://terinstock.com/post/2018/10/memfd_create-Temporary-in-memory-files-with-Go-and-Linux/) via @zbuc. I also don't fully understand the pattern, but it does indeed work..

pkg/cmd/secrets_edit.go
apazzolini: I tested, and other processes running as the same user can still access the file. Chris was going to research the flags to see if we can lock it down further, but I'm not sure if he's had a chance to do so yet. I think it's still better than writing to disk though.

apazzolini: Discussed offline 👍

pkg/cmd/secrets_edit_linux_amd64.go
```diff
@@ -0,0 +1,226 @@
+/*
+Copyright © 2020 Doppler <support@doppler.com>
```
ScottBrenner, line 2:
```suggestion
Copyright © 2022 Doppler <support@doppler.com>
```

apazzolini:
Closing - we won't be doing this.

PR #264, opened 2021-10-06
Add env to cli-action

jlarmstrongiv:
If you installed the doppler cli in github actions, then more often than not you will need a DOPPLER_TOKEN in the environment

### DIFF
--- a/.github/workflows/cli-action.yml
+++ b/.github/workflows/cli-action.yml
@@ -10,6 +10,8 @@ jobs:
       uses: dopplerhq/cli-action@v1
     - name: Test CLI
       run: doppler --version
+      env:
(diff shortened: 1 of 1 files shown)


Piccirello:
Could you provide some more info on what this accomplishes? The token isn't used for displaying the CLI version.

jlarmstrongiv:
Maybe a comment instead?  Or, this can be closed.

When using the cli for anything significant, you’ll likely need a DOPPLER_TOKEN in the environment.  It wasn’t clear to me that I needed to set that manually.  A note about it could save an hour of debugging

Piccirello:
Thanks for that info, I see what you mean now. I just added [this](https://github.com/DopplerHQ/cli-action#configuration) section to our action's README which should hopefully help future users.

PR #268, opened 2021-10-14
Maintain local config file ownership when modifying config

Piccirello:
This fixes an issue where running `sudo doppler update` would result in the user's config file being owned by root. I tested an implementation locally that instead modifies the existing file in-place (or creates it if it doesn't exist), but I couldn't find a way to perform the [update or create] operation atomically.

Closes ENG-3389.

### DIFF
--- a/pkg/utils/io_nonwindows.go
+++ b/pkg/utils/io_nonwindows.go
@@ -1,0 +1,38 @@
+// +build !windows
+
+/*
+Copyright © 2021 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/utils/io_nonwindows.go
```diff
@@ -0,0 +1,38 @@
...
+package utils
+
+import (
+	"errors"
+	"os"
+	"syscall"
+)
+
+func FileOwnership(path string) (int, int, error) {
+	info, err := os.Stat(path)
```
nmanoogian, line 27: Is this `Stat` necessary if we're going to do the syscall stat below?

nmanoogian (approved):

rgmvisser (approved):

pkg/utils/io_nonwindows.go
Piccirello: This `Stat()` produces an `info` object, when we then call `Sys()` on below, before coercing it to the type `*syscall.Stat_t`. It's definitely not the cleanest API, but this is how Go goes it 🤷

PR #270, opened 2021-10-22
Replace keychain library on macOS

Piccirello:
Closes ENG-3314.


Piccirello:
As much as I really like this new library, it requires cgo, which goreleaser does not play very nicely with. This is going to be more effort than we can expend on this rn.

PR #272, opened 2021-10-23
Support supplying `doppler configure set` value on stdin

Piccirello:
Closes ENG-3346.


nmanoogian (approved):
Looking good! I did some light local testing with this and everything worked well for me.

tests/e2e/configure.sh
```diff
@@ -101,6 +101,13 @@ config="$("$DOPPLER_BINARY" configure get config --configuration=./temp-config -
 
 beforeEach
 
+# test set using stdin
```
rgmvisser, line 104: Nice!

PR #276, opened 2021-10-25
Add support for doppler completion installation for zsh

Kunamatata:
Closes ENG-2640

### DIFF
--- a/pkg/cmd/completion.go
+++ b/pkg/cmd/completion.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -47,7 +48,7 @@ var completionCmd = &cobra.Command{
 				utils.HandleError(err, "Unable to generate fish completions.")
 			}
 		} else {
-			utils.HandleError(fmt.Errorf("Your shell is not supported"))
+			utils.HandleError(errors.New("Your shell is not supported"))
 		}
 	},
 }
@@ -72,6 +73,11 @@ var completionInstallCmd = &cobra.Command{
 		var buf bytes.Buffer
 		var path string
 		var name string
+
+		if utils.IsWindows() {
+			utils.HandleError(errors.New("Completion files are not supported on Windows. You can use completion files with Windows Subsystem for Linux (WSL)."))
+		}
+
 		if strings.HasSuffix(shell, "/bash") {
 			if err := cmd.Root().GenBashCompletion(&buf); err != nil {
 				utils.HandleError(err, "Unable to generate bash completions.")
@@ -82,8 +88,15 @@ var completionInstallCmd = &cobra.Command{
 			} else {
 				path = "/etc/bash_completion.d"
 			}
+		} else if strings.HasSuffix(shell, "/zsh") {
+			if err := cmd.Root().GenZshCompletion(&buf); err != nil {
+				utils.HandleError(err, "Unable to generate zsh completions.")
+			}
+			// zsh completions file start with an underscore
+			name = "_doppler"
+			path = "/usr/local/share/zsh/site-functions"
 		} else {
-			utils.HandleError(fmt.Errorf("Your shell is not supported"))
+			utils.HandleError(errors.New("Your shell is not supported"))
 		}
 
 		// create directory if it doesn't exist
@@ -117,7 +130,7 @@ func getShell(args []string) string {
 		shell = fmt.Sprintf("%s", args[0])
 	}
 	if shell == "" {
-		utils.HandleError(fmt.Errorf("Unable to determine current shell"), "Please provide your shell name as an argument")
+		utils.HandleError(errors.New("Unable to determine current shell"), "Please provide your shell name as an argument")
 	}
 
 	// normalize shell


pkg/cmd/completion.go
```diff
@@ -82,6 +83,15 @@ var completionInstallCmd = &cobra.Command{
...
 				path = "/etc/bash_completion.d"
 			}
+		} else if strings.HasSuffix(shell, "/zsh") {
+			if err := cmd.Root().GenZshCompletion(&buf); err != nil {
+				utils.HandleError(err, "Unable to generate zsh completions.")
+			}
+			// zsh completions file start with an underscore
+			name = "_doppler"
+			if utils.IsMacOS() {
+				path = "/usr/local/share/zsh/site-functions"
```
Kunamatata: After doing some research this folder seems like the best place to put the `_doppler` completion file.
However, we could also modify the `$fpath` variable and insert a custom directory (for example: ~/.zsh/completions/doppler/...` or some variation).

zsh looks in `$fpath` to find completion files and functions:
> The convention for autoloaded functions used in completion is that they start with an underscore; as already mentioned, the fpath/FPATH parameter must contain the directory in which they are stored. If zsh was properly installed on your system, then fpath/FPATH automatically contains the required directories for the standard functions.

Piccirello (approved):
Very nice! Pro tip: you can reference the corresponding linear ticket in the PR description using "Closes ENG-2640". That'll auto-link the tickets and mark it resolved once the PR is merged.

pkg/cmd/completion.go
```diff
@@ -82,6 +83,15 @@ var completionInstallCmd = &cobra.Command{
...
+			}
```
Piccirello: How does this handle being run on a different OS? It looks like we never set the path.

Piccirello (changes requested):

pkg/cmd/completion.go
Kunamatata: You are correct, I believe we don't need the check anymore. Seems like zsh installs in the same directories for different OS (unix based)

I installed `zsh` on my windows via WSL and the path was the same `/usr/local/share/zsh/site-functions`

Not sure about native windows without WSL.

pkg/cmd/completion.go
```diff
@@ -89,7 +89,7 @@ var completionInstallCmd = &cobra.Command{
 			}
 			// zsh completions file start with an underscore
 			name = "_doppler"
-			if utils.IsMacOS() {
+			if !utils.IsWindows() {
```
Kunamatata: If Windows users use the "Windows Subsystem for Linux" the path for zsh remains the same.

Piccirello: Should we exit with an error when using Windows? The below code doesn't appear to handle `path` being nil.

Piccirello:
Can you squash all of your commits into one?

pkg/cmd/completion.go
Kunamatata: What do you think of adding the following right underneath the variable declarations of the function and removing the `if !utils.isWindows()` check from inside the `zsh` check?
```go
if utils.IsWindows() {
	utils.Log("Completion files are not supported on windows currently. However you can use completion files with Windows Subsystem for Linux.")
	return
}
```

Piccirello: I like that since it's helpful to the user. We'll want to exit with an error since installation failed.

pkg/cmd/completion.go
```diff
@@ -72,6 +72,11 @@ var completionInstallCmd = &cobra.Command{
 		var buf bytes.Buffer
 		var path string
 		var name string
+
+		if utils.IsWindows() {
+			utils.HandleError(fmt.Errorf("Completion files are not supported on windows currently. However you can use completion files with Windows Subsystem for Linux."))
+		}
```
Piccirello: Is this true for bash too? We may just want to guard this in the zsh section.

Kunamatata: It is true for bash too. Windows native does not have bash. 
However if they use WSL they will be able to use bash and the setup will continue since `utils.IsWindows()` will return false on a WSL distro.

pkg/cmd/completion.go
```diff
@@ -72,6 +72,11 @@ var completionInstallCmd = &cobra.Command{
 		var buf bytes.Buffer
 		var path string
 		var name string
+
+		if utils.IsWindows() {
+			utils.HandleError(fmt.Errorf("Completion files are not supported on windows currently. However you can use completion files with Windows Subsystem for Linux."))
```
Piccirello: nit: you can use `errors.New` here since we don't need a template string

```suggestion
			utils.HandleError(errors.New("Completion files are not supported on windows currently. However you can use completion files with Windows Subsystem for Linux."))
```

Kunamatata: Resolved by 25e04da

pkg/cmd/completion.go
```diff
@@ -72,6 +73,11 @@ var completionInstallCmd = &cobra.Command{
 		var buf bytes.Buffer
 		var path string
 		var name string
+
+		if utils.IsWindows() {
+			utils.HandleError(errors.New("Completion files are not supported on windows currently. However you can use completion files with Windows Subsystem for Linux."))
```
Piccirello:
```suggestion
			utils.HandleError(errors.New("Completion files are not supported on Windows. You can use completion files with Windows Subsystem for Linux (WSL)."))
```

Piccirello (commented):
One last nit: the changelog for this repo is autogenerated by the commit messages. Given that this repo is exclusively for the CLI, you can remove the mention of the CLI from it.

"Add zsh completion support for doppler cli" -> "Add support for installing zsh completions"

Piccirello (approved):
Congrats- you're now a CLI contributor 😃

PR #278, opened 2021-11-03
chore: fix num args expected for 'doppler import'

Piccirello:
This function doesn't use args.

### DIFF
--- a/pkg/cmd/import.go
+++ b/pkg/cmd/import.go
@@ -32,7 +32,7 @@ const projectTemplateFileName = "doppler-template.yaml"
 var importCommand = &cobra.Command{
 	Use:   "import",
 	Short: "Import projects into your Doppler workplace",
-	Args:  cobra.MaximumNArgs(1),
+	Args:  cobra.NoArgs,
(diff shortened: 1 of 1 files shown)


pkg/cmd/import.go
```diff
@@ -32,7 +32,7 @@ const projectTemplateFileName = "doppler-template.yaml"
 var importCommand = &cobra.Command{
 	Use:   "import",
 	Short: "Import projects into your Doppler workplace",
-	Args:  cobra.MaximumNArgs(1),
+	Args:  cobra.NoArgs,
```
rgmvisser, line 35: This has a template flag where you can specify what file to use (but I guess this is a flag and not an arg?).

rgmvisser (approved):

nmanoogian (approved):

pkg/cmd/import.go
Piccirello: Flags are handled a bit differently than args. Ideally this command should probably accept the template path via flag or via arg (like our other commands do- e.g. `doppler projects create`).

PR #280, opened 2021-11-05
Support installing CLI to custom path

Piccirello:
This PR adds support for installing the CLI to a custom path via the install.sh script. This feature will enable the CLI to someday support updating itself when running out of a custom directory.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -14,6 +14,8 @@ USE_PACKAGE_MANAGER=1
 VERIFY_SIGNATURE=1
 FORCE_VERIFY_SIGNATURE=0
 DISABLE_CURL=0
+CUSTOM_INSTALL_PATH=""
+BINARY_INSTALLED_PATH=""
 
 tempdir=""
 filename=""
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -529,6 +529,23 @@ elif [ "$format" = "tar" ]; then
       fi
     fi
 
+    # support creating /usr/local/bin if it's in the PATH but doesn't exist.
+    # this fixes an issue with clean installs of macOS 12+
+    install_dir="/usr/local/bin"
+    if [ ! -e "$install_dir" ] && [ "$( is_dir_in_path "$install_dir" )" -eq "1" ]; then
+      log_debug "/usr/local/bin is in PATH but doesn't exist"
+      log_debug "Creating /usr/local/bin"
+      if mkdir -m 755 "$install_dir" > /dev/null 2>&1 ; then
```
nmanoogian: Would it be worthwhile to add the `-p` flag as well, in the event that the parent dirs also don't exist?

nmanoogian (dismissed):

scripts/install.sh
Piccirello: I don't think so- if `/usr/local/` doesn't exist there's a bigger problem.

nmanoogian (dismissed):

Piccirello:
This PR isn't quite ready yet- there's a shellcheck failure I need to fix. I've also split it up so I can ship the other commit more quickly (#281). Will take this out of draft mode once it's ready for another review.

nmanoogian (approved):

PR #281, opened 2021-11-05
Support creating /usr/local/bin during install

Piccirello:
We only create the directory if it's already in the PATH.


rgmvisser (commented):
Looks good from my side, but will let @nmanoogian stamp it

nmanoogian (approved):

PR #284, opened 2021-11-16
Log informational messages to stderr

### LINKED ISSUE #271: [BUG] the info message printed with fallback should be sent to stderr, not stdout
**Describe the bug**
I was trying to do
```
myvar="$(doppler run --fallback-only -- some command here)"
```
...expecting to just get the output of `some command here`, however the output `Reading secrets from fallback file` is also part of the value of `myvar`

**To Reproduce**

**Expected behavior**
Only after experiencing this error did I discover the `--silent` flag which works to eliminate this message. I think it's pretty standard for such messages to go directly to stderr instead of stdout. What do you think?

**Screenshots**

 - OS: alpine
 - Version latest

**CLI Version:**
Version 3.34.1

**Additional context**

Piccirello:
```
$ ./scripts/install.sh 
Downloading Doppler CLI
Verifying signature
Installing...
Installed Doppler CLI v3.35.0
$ ./scripts/install.sh 2>/dev/null
Installed Doppler CLI v3.35.0
$ ./scripts/install.sh --debug 2>/dev/null
Installed Doppler CLI v3.35.0
```

Closes ENG-3575, closes #271.

### DIFF
--- a/pkg/utils/log.go
+++ b/pkg/utils/log.go
@@ -23,25 +23,29 @@ import (
 	"gopkg.in/gookit/color.v1"
 )
 
-// Log info message to stdout
+// Print output to stdout
+func Print(info string) {
+	fmt.Println(info)
+}
+
+// Print output to stdout.
+func PrintWarning(s string) {
+	fmt.Println(color.Yellow.Render("Warning:"), s)
+}
+
+// Log info message to stderr
 func Log(info string) {
-	if CanLogInfo() {
-		fmt.Println(info)
-	}
+	fmt.Fprintln(os.Stderr, info)
 }
 
 // LogWarning message to stdout
 func LogWarning(s string) {
-	if CanLogInfo() {
-		fmt.Println(color.Yellow.Render("Warning:"), s)
-	}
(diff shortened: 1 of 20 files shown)


pkg/utils/log.go
```diff
@@ -23,25 +23,29 @@ import (
...
+}
+
+// Print output to stdout.
+func PrintWarning(s string) {
+	fmt.Println(color.Yellow.Render("Warning:"), s)
+}
+
+// Log info message to stderr
 func Log(info string) {
-	if CanLogInfo() {
```
rgmvisser, line 28: This is not needed anymore?

rgmvisser (approved):

nmanoogian (approved):

pkg/utils/log.go
Piccirello: Since we're now logging these messages to stderr, we don't need to worry about adding noise to the output (which is on stdout).

rgmvisser: Gotcha, should we have these checks in `func Print(info string) {` and `func PrintWarning(s string) {` then as they are now stdout?

PR #288, opened 2021-11-24
Shiftleft action config 1637796310

kdsjfl:


Piccirello:
Can you provide some info on the purpose of this?

PR #290, opened 2021-12-29
Add support for mounting secrets to file path

Piccirello:
Currently, `doppler run` always inject secrets as environment variables. With this PR, secrets can now instead be mounted to an ephemeral file. The file exists for the lifetime of the application and can be read by libraries like dotenv. The path to the file is made available in the `DOPPLER_CLI_SECRETS_PATH` environment variable.

This feature supports custom name transformers.

Example:
```sh
$ doppler run --mount secrets.json -- sh
# new shell
$ echo $DOPPLER_CLI_SECRETS_PATH
/Users/doppler/.doppler/secrets.json
$ cat $DOPPLER_CLI_SECRETS_PATH
'{"TEST_SECRET":"hello"}'
```
Example w/ Node:
```node
$ doppler run --mount secrets.json -- node
> fs.readFileSync(process.env.DOPPLER_CLI_SECRETS_PATH).toString()
'{"TEST_SECRET":"hello"}'
$ doppler run --mount .env -- node
> process.env
{
  DOPPLER_CLI_SECRETS_PATH: '/usr/src/app/.env'
}
> require('dotenv').config()
> process.env
{
  DOPPLER_CLI_SECRETS_PATH: '/usr/src/app/.env',
  TEST_SECRET: 'hello'
}
```

New flags for `doppler run`:
| Flag        | Default           | Description  |
| ------------- |-------------| -----|
|`mount` | (off) | path to mount secrets file to, accessible via DOPPLER_CLI_SECRETS_PATH env var |
| `mount-format` | json | file format to use. one of [json, env] |
| `mount-max-reads` | 0 | maximum number of times the mounted secrets file can be read (0 for unlimited) |

## More info

Technically, the mounted secrets file is a named pipe. This allows us to ensure that the pipe's contents (i.e. the secrets) are only accessible while the CLI is connected to it. Once the CLI exits, the pipe is useless.

Potential gotchas:
- Next.js appears to [disallow reading from FIFOs](https://github.com/vercel/next.js/blob/636d3aeda753301657cf9f2e9faa8e947392bfe2/packages/next-env/index.ts#L99-L102), other libraries might too
- After writing the secrets to the FIFO, we close the FIFO to signal to the reader that we're finished. We then wait 10 ms before re-opening the FIFO for further reads. This delay seems arbitrary but, in testing, is required for the reader to be able to detect the EOF. Of  course the 10ms number itself is arbitrary and may need to be adjusted at some point.

Closes ENG-3665

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -40,6 +40,16 @@ var defaultFallbackDir string
 
 const defaultFallbackFileMaxAge = 14 * 24 * time.Hour // 14 days
 
+type fallbackOptions struct {
+	enable             bool
+	path               string
+	legacyPath         string
+	readonly           bool
+	exclusive          bool
+	exitOnWriteFailure bool
+	passphrase         string
+}
+
 var runCmd = &cobra.Command{
 	Use:   "run [command]",
 	Short: "Run a command with secrets injected into the environment",
@@ -112,7 +122,16 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
 			}
 		}
 
-		secrets := fetchSecrets(localConfig, enableCache, enableFallback, fallbackPath, legacyFallbackPath, metadataPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, passphrase, nameTransformer, dynamicSecretsTTL)
+		fallbackOpts := fallbackOptions{
+			enable:             enableFallback,
+			path:               fallbackPath,
+			legacyPath:         legacyFallbackPath,
+			readonly:           fallbackReadonly,
+			exclusive:          fallbackOnly,
+			exitOnWriteFailure: exitOnWriteFailure,
+			passphrase:         passphrase,
+		}
+		secrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL)
 
 		if preserveEnv {
 			utils.LogWarning("Ignoring Doppler secrets already defined in the environment due to --preserve-env flag")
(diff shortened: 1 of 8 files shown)


pkg/cmd/run.go
```diff
@@ -111,41 +121,89 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
-				env = append(env, fmt.Sprintf("%s=%s", name, value))
+			absMountPath, handler, err := controllers.MountSecrets(secrets, mountFormat, mountPath, maxReads)
+			if !err.IsNil() {
+				utils.HandleError(err.Unwrap(), err.Message)
+			}
+			mountPath = absMountPath
+			onExit = handler
+
+			// export path to mounted file
+			env = append(env, fmt.Sprintf("%s=%s", "DOPPLER_CLI_SECRETS_PATH", mountPath))
```
ryan-blunden, line 216: Suggestion to go with "mount" over "secrets" and condense the var name.

```suggestion
			env = append(env, fmt.Sprintf("%s=%s", "DOPPLER_MOUNT_PATH", mountPath))
```

pkg/cmd/run.go
```diff
@@ -570,6 +628,10 @@ func init() {
 	runCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
 	runCmd.Flags().Bool("no-exit-on-write-failure", false, "do not exit if unable to write the fallback file")
 	runCmd.Flags().Bool("forward-signals", forwardSignals, "forward signals to the child process (defaults to false when STDOUT is a TTY)")
+	// secrets mount flags
+	runCmd.Flags().String("mount", "", "write secrets to an ephemeral file, accessible at DOPPLER_CLI_SECRETS_PATH. when enabled, secrets are NOT injected into the environment")
+	runCmd.Flags().String("mount-format", "json", "file format to use. one of [json, env]")
+	runCmd.Flags().Int("mount-max-reads", 0, "maximum number of times the mounted secrets file can be read (0 for unlimited)")
```
ryan-blunden, line 652:
```suggestion
	runCmd.Flags().Int("mount-max-reads", 0, "number of read requests permitted (0 for unlimited)")
```

pkg/cmd/run.go
```diff
@@ -570,6 +628,10 @@ func init() {
 	runCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
 	runCmd.Flags().Bool("no-exit-on-write-failure", false, "do not exit if unable to write the fallback file")
 	runCmd.Flags().Bool("forward-signals", forwardSignals, "forward signals to the child process (defaults to false when STDOUT is a TTY)")
+	// secrets mount flags
+	runCmd.Flags().String("mount", "", "write secrets to an ephemeral file, accessible at DOPPLER_CLI_SECRETS_PATH. when enabled, secrets are NOT injected into the environment")
```
ryan-blunden, line 650: Might be worth stating that the CLI automatically sets the format if a file extension is provided as well as calling out the name transformers are supported would be beneficial as they will negate the need for the `--mount-format` option in nearly all cases.

```suggestion
	runCmd.Flags().String("mount", "", mount ephemeral secrets file in format set by file extension (JSON by default). supports --name-transformers. no environment variables injected except $DOPPLER_MOUNT_PATH.")
```

ryan-blunden (commented):
Could we add a usage example, e.g. `doppler run --mount .env -- start app`

pkg/cmd/run.go
Piccirello: I like the change from `SECRETS` -> `MOUNT`. I'm leaning towards keeping the `CLI` prefix so that this variable is namespaced properly. There's currently nothing stopping the user from having a `DOPPLER_MOUNT_PATH` secret in their config, but we do prevent users from creating `DOPPLER_CLI` secrets.

pkg/cmd/run.go
Piccirello: Do you think the new language makes it clear what this feature does? I wonder if it's confusing from the perspective of a user who knows nothing about named pipes.

ryan-blunden: I think the new language is clearer and more concise and the usage of named pipes is an implementation detail that doesn't benefit the user to know for general usage.

pkg/cmd/run.go
ryan-blunden: Ok got it. Whatever you think is best.

Piccirello:
Comments have been addressed. The env format now wraps all values in quotes. I'm going to push us pretty hard to limit the number of supported formats at launch, as well as any other features that are a nice to have. We can add more functionality over time.

ryan-blunden (dismissed):
LGTM!

pkg/cmd/run.go
```diff
@@ -131,52 +133,97 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
 					useSecret = false
 					break
 				}
 			}
+			if !useSecret {
+				continue
+			}
 
-			if useSecret && preserveEnv {
+			if preserveEnv && existingEnvKeys[name] == true {
```
nmanoogian: Should we explicitly disallow the use of `--preserve-env` with `--mount`? Or does it still make sense to make that option available?

As it stands, there's some weird behavior where if you have a secret called `MY_SECRET` and you do `MY_SECRET=123 doppler run --mount secrets.json --preserve-env -- cat secrets.json`, you'll get a `secrets.json` which is missing `MY_SECRET`.

pkg/cmd/run.go
```diff
@@ -131,52 +133,97 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
 			key := parts[0]
 			existingEnvKeys[key] = true
 		}
 
+		secrets := map[string]string{}
 		excludedKeys := []string{"PATH", "PS1", "HOME"}
-		for name, value := range secrets {
+		for name, value := range dopplerSecrets {
 			useSecret := true
 			for _, excludedKey := range excludedKeys {
```
nmanoogian: Maybe we should only exclude these keys if `--mount` is not specified?

nmanoogian (dismissed):
Super clean implementation 👍 It feels like we should have done this ages ago!

pkg/cmd/run.go
Piccirello: Good catch

nmanoogian (approved):
Looks good! I haven't had a chance to play with the latest changes because of the issue I posted about in #cli

pkg/cmd/run.go
Piccirello: Added a test for this!

pkg/cmd/run.go
```diff
@@ -112,52 +124,114 @@ doppler run --command "YOUR_COMMAND && YOUR_OTHER_COMMAND"`,
...
+						useSecret = false
+						break
+					}
 				}
-			}
 
-			if useSecret && preserveEnv {
 				// skip secret if environment already contains variable w/ same name
-				if existingEnvKeys[name] == true {
+				if preserveEnv && existingEnvKeys[name] == true {
```
nmanoogian, line 178: Should this logic be flipped? If we're not mounting the file, we want to set `useSecret = false` when `preserveEnv == false` and there is an existing key matching.

nmanoogian (changes requested):

pkg/cmd/run.go
nmanoogian: Hold on a minute, I've gone and confused myself. Please stand by

nmanoogian: Disregard, this logic is correct 👍

nmanoogian (approved):

PR #292, opened 2022-01-03
Various minor install.sh improvements

Piccirello:
The individual commits are self-explanatory.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -194,7 +194,7 @@ wget_download() {
   exit_code=$?
   set -e
 
-  status_code="$(echo "$headers" | sed '1!G;h;$!d' | grep -o -E 'HTTP/[0-9.]+ [0-9]{3}' | head -1 | grep -o -E '[0-9]{3}')"
+  status_code="$(echo "$headers" | grep -o -E '^\s*HTTP/[0-9.]+ [0-9]{3}' | tail -1 | grep -o -E '[0-9]{3}')"
   # it's possible for this value to be blank, so confirm that it's a valid status code
   valid_status_code=0
   if expr "$status_code" : '[0-9][0-9][0-9]$'>/dev/null; then
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -221,7 +221,9 @@ check_http_status() {
 is_dir_in_path() {
   dir="$1"
   found=1
-  (echo "$PATH" | grep -q "$dir:" || echo "$PATH" | grep -q -E "$dir$") || found=0
+  # ensure dir is the full path and not a substring of some longer path.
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -221,7 +221,8 @@ check_http_status() {
 is_dir_in_path() {
   dir="$1"
   found=1
-  (echo "$PATH" | grep -q "$dir:" || echo "$PATH" | grep -q -E "$dir$") || found=0
+  # ensure dir is the full path and not a substring of some longer path
+  ( (echo "$PATH" | grep -q -E "(^|:)$dir(:|$)") || (echo "$PATH" | grep -q -E "$dir$") ) || found=0
```
nmanoogian: It looks like we're using regex here already but are there any concerns about having special characters in the path? `[]` for example

scripts/install.sh
```diff
@@ -71,7 +71,7 @@ delete_tempdir() {
 
 linux_shell() {
   user="$(whoami)"
-  grep "$user" < /etc/passwd | cut -f 7 -d ":" | head -1
+  grep -E "^$user" < /etc/passwd | cut -f 7 -d ":" | head -1
```
nmanoogian: Same question here on regex special chars. Additionally, should this be `"^$user:" to make sure we don't match a user who is a substring of another user?

nmanoogian (dismissed):
Awesome improvements!

scripts/install.sh
Piccirello: Whoops, I definitely intended to include the colon in this change. Good catch.

scripts/install.sh
Piccirello: I believe this would already handle special characters like that. Am I overlooking something glaring?

nmanoogian: The one I'm mainly worried about is `.` -- For example, `foo.biz` would match `foolbiz` or `foo/biz`

Piccirello: Special characters like `.` could indeed result in an erroneous match. Practically, we don't need to worry about this today because the `is_dir_in_path` check is only performed against a static list of paths (`/usr/local/bin`, `/usr/bin`, `/usr/sbin`). I'll add a comment to the function noting this.

Piccirello: I ended up being able to clean this up a bit by performing a fixed-string (i.e. non-regex) search against the regex search results. This should prevent most special characters from getting in the way, unless they cause the initial regex search to miss results.

https://github.com/DopplerHQ/cli/compare/7d25f66544f587f77c58af09535f82dad222f564..295f28d628aa344342d12ef17fe70419290686e5?w=1

nmanoogian (approved):

PR #295, opened 2022-01-20
Add support for dynamic secrets in standard secret commands

nmanoogian:
- `doppler secrets`, `doppler secrets download`, `doppler secrets substitute`, `doppler run`
  - Will issue and return dynamic secrets by default, if they exist
  - Accepts optional `--dynamics-expire <duration>`, otherwise uses system default 
  - `doppler secrets --only-names` will not issue any dynamic secrets
- `doppler secrets get <secret_name>`
  - Will not issue dynamic secrets, by design

### DIFF
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -165,7 +165,7 @@ func secrets(cmd *cobra.Command, args []string) {
 
 	utils.RequireValue("token", localConfig.Token.Value)
 
-	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, nil, false, 0)
 	if !err.IsNil() {
 		utils.HandleError(err.Unwrap(), err.Message)
 	}
@@ -191,7 +191,11 @@ func getSecrets(cmd *cobra.Command, args []string) {
 
 	utils.RequireValue("token", localConfig.Token.Value)
 
-	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+	var requestedSecrets []string
+	if len(args) > 0 {
+		requestedSecrets = args
+	}
+	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, requestedSecrets, false, 0)
 	if !err.IsNil() {
 		utils.HandleError(err.Unwrap(), err.Message)
 	}
@@ -394,6 +398,7 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
 	fallbackReadonly := utils.GetBoolFlag(cmd, "fallback-readonly")
 	fallbackOnly := utils.GetBoolFlag(cmd, "fallback-only")
 	exitOnWriteFailure := !utils.GetBoolFlag(cmd, "no-exit-on-write-failure")
+	dynamicSecretsExpire := utils.GetDurationFlag(cmd, "expire-dynamics")
 
 	utils.RequireValue("token", localConfig.Token.Value)
 
@@ -444,7 +449,7 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
 		if enableCache {
 			metadataPath = controllers.MetadataFilePath(localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
 		}
-		secrets := fetchSecrets(localConfig, enableCache, enableFallback, fallbackPath, legacyFallbackPath, metadataPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, fallbackPassphrase, nameTransformer)
+		secrets := fetchSecrets(localConfig, enableCache, enableFallback, fallbackPath, legacyFallbackPath, metadataPath, fallbackReadonly, fallbackOnly, exitOnWriteFailure, fallbackPassphrase, nameTransformer, dynamicSecretsExpire)
 
 		var err error
 		body, err = json.Marshal(secrets)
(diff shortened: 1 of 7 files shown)


pkg/cmd/secrets.go
```diff
@@ -647,6 +657,7 @@ func init() {
 	secretsSubstituteCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
 	secretsSubstituteCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	secretsSubstituteCmd.Flags().String("output", "", "path to the output file. by default the rendered text will be written to stdout.")
+	secretsSubstituteCmd.Flags().Duration("dynamics-expire", 0, "dynamic secrets will expire after specified duration, (e.g. '3h', '15m')")
```
Piccirello: You'll want to add these flags to the deprecated `enclave secrets` commands too

pkg/cmd/secrets.go
```diff
@@ -634,6 +643,7 @@ func init() {
 	secretsDownloadCmd.Flags().String("name-transformer", "", fmt.Sprintf("(BETA) output name transformer. one of %v", validNameTransformersList))
 	secretsDownloadCmd.Flags().String("passphrase", "", "passphrase to use for encrypting the secrets file. the default passphrase is computed using your current configuration.")
 	secretsDownloadCmd.Flags().Bool("no-file", false, "print the response to stdout")
+	secretsDownloadCmd.Flags().Duration("dynamics-expire", 0, "dynamic secrets will expire after specified duration, (e.g. '3h', '15m')")
```
Piccirello: Does it make sense for this to be a runtime option? As opposed to using whatever the dynamic has been configured for. (if we do keep it, I think `--expire-dynamics` sounds better)

pkg/http/api.go
```diff
@@ -159,11 +167,28 @@ func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string,
...
+	if secrets != nil {
+		params = append(params, queryParam{Key: "secrets", Value: strings.Join(secrets, ",")})
+	}
+
+	var includeDynamicSecretsOption string
+	if includeDynamicSecrets {
+		includeDynamicSecretsOption = "true"
+	} else {
+		includeDynamicSecretsOption = "false"
+	}
```
Piccirello: super nit
```suggestion
	includeDynamicSecretsOption := "false"
	if includeDynamicSecrets {
		includeDynamicSecretsOption = "true"
	}
```

pkg/http/api.go
```diff
@@ -159,11 +167,28 @@ func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string,
...
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "config", Value: config})
 
+	if secrets != nil {
+		params = append(params, queryParam{Key: "secrets", Value: strings.Join(secrets, ",")})
+	}
+
+	var includeDynamicSecretsOption string
+	if includeDynamicSecrets {
+		includeDynamicSecretsOption = "true"
```
Piccirello, line 181: Can we use booleans here or is the string intentional?

pkg/cmd/secrets.go
nmanoogian: I think long-term, we'll allow users to set a max and default during setup of the dynamic secrets. I think there is some value in letting you set it on the CLI for context-specific reasons. For example, I could imagine an AWS user that one CI job will only need for 5 min max (restart service) vs 1 hour max (large bundle upload).

pkg/http/api.go
nmanoogian: It just needs to be string because it's a query param (GET request)

pkg/cmd/secrets.go
```diff
@@ -165,7 +165,9 @@ func secrets(cmd *cobra.Command, args []string) {
 
 	utils.RequireValue("token", localConfig.Token.Value)
 
-	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value)
+	includeDynamicSecrets := !onlyNames
+	dynamicSecretsExpire := utils.GetDurationFlag(cmd, "dynamics-expire")
+	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, nil, includeDynamicSecrets, dynamicSecretsExpire)
```
Piccirello: Discussed offline: we won't support dynamics in the `secrets` command

pkg/cmd/secrets.go
Piccirello: Wondering if this would be more appropriate as a follow up item, once the use case has established itself a bit more. Maybe it's already clear that we need this though. Will defer to you.

nmanoogian: Good looking out -- I do think we've got some clear use cases for this. My gut says ship it

nmanoogian:
Force push addresses feedback and adds smokecheck e2e tests for legacy `doppler enclave` commands

tests/e2e/legacy-commands.sh
```diff
@@ -0,0 +1,68 @@
...
+
+beforeEach
+
+# test 'doppler enclave secrets'
+"$DOPPLER_BINARY" enclave secrets || (echo "ERROR: doppler enclave secrets" && exit 1)
+
+beforeEach
+
+# test 'doppler enclave secrets download'
+"$DOPPLER_BINARY" enclave secrets download || (echo "ERROR: doppler enclave secrets download" && exit 1)
```
Piccirello: You'll want to manually specify (and clean up) the file that's generated

pkg/http/api.go
```diff
@@ -136,11 +138,17 @@ func RevokeAuthToken(host string, verifyTLS bool, token string) (map[string]inte
...
+func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string, config string, format models.SecretsFormat, nameTransformer *models.SecretsNameTransformer, etag string, dynamicSecretsExpire time.Duration) (int, http.Header, []byte, Error) {
 	var params []queryParam
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "config", Value: config})
 	params = append(params, queryParam{Key: "format", Value: format.String()})
+	params = append(params, queryParam{Key: "include_dynamic_secrets", Value: "true"})
+
+	if dynamicSecretsExpire > 0 {
+		ttlSeconds := int(dynamicSecretsExpire.Seconds())
+		params = append(params, queryParam{Key: "dynamic_secrets_ttl", Value: strconv.Itoa(ttlSeconds)})
```
Piccirello: Should we include the unit in this key? So `dynamic_secrets_ttl_sec`

Piccirello (dismissed):

nmanoogian:
Sorry @Piccirello, my urge for consistency got the better of me. That last force push just fixed a commit message but it still cleared your approval

Piccirello (approved):

PR #298, opened 2022-02-07
[install.sh]: require signature verification

Piccirello:
This should be merged on February 9th.


nmanoogian (dismissed):

Piccirello:
Removed some other conditional logic that I had missed.

nmanoogian (approved):

PR #299, opened 2022-02-09
Surface gpg error without --debug and fix typo

ryan-blunden:
Given that some folks are going to run into issues with `doppler update` due to gpg now being required, I think it's worth surfacing this error without `--debug`.

The error output would now be:

```sh
doppler update
Updating...
Error: Unable to find gpg binary for signature verification. You can resolve this error by installing your system's gnupg package
Doppler Error: exit status 1
```

Also fixed a typo in install.sh

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -69,12 +69,17 @@ func RunInstallScript() (bool, string, Error) {
 		exitCode := 1
 		if exitError, ok := err.(*exec.ExitError); ok {
 			exitCode = exitError.ExitCode()
-		}
+		}		
 
 		message := "Unable to install the latest Doppler CLI"
 		permissionError := exitCode == 2 || strings.Contains(strOut, "dpkg: error: requested operation requires superuser privilege")
 		if permissionError {
 			message = "Error: update failed due to improper permissions\nPlease re-run with `sudo` or as an admin"
+		}
+
+		gpgError := strings.Contains(strOut, "Unable to find gpg")
+		if gpgError {
+			message = "Error: Unable to find gpg binary for signature verification\nYou can resolve this error by installing your system's gnupg package"
 		}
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/update.go
```diff
@@ -69,14 +69,19 @@ func RunInstallScript() (bool, string, Error) {
...
-		}
+		}		
 
 		message := "Unable to install the latest Doppler CLI"
 		permissionError := exitCode == 2 || strings.Contains(strOut, "dpkg: error: requested operation requires superuser privilege")
 		if permissionError {
 			message = "Error: update failed due to improper permissions\nPlease re-run with `sudo` or as an admin"
 		}
 
+		gpgError := strings.Contains(strOut, "Unable to find gpg")
```
Piccirello, line 80: You'll want to use a unique exit code for this scenario, like the permission error above

scripts/install.sh
```diff
@@ -378,7 +378,7 @@ gpg_binary="$(command -v gpg || true)";
 if [ -x "$gpg_binary" ]; then
   log_debug "Using $gpg_binary for signature verification"
 else
-  log "ERROR: Unable to find gpg binary for signature verficiation"
+  log "ERROR: Unable to find gpg binary for signature verification"
```
Piccirello, line 381: Can you move this to its own commit? You'll want to prepend it it with `chore:`

Piccirello (commented):
The interesting thing about this scenario is that pulling in this code will require updating the CLI. That won't be possible if `gpg` isn't available. If this new CLI version with this code *is* installed, then we also know that `gpg` was available at install time. So I'm not sure how meaningful this change will actually be.

pkg/controllers/update.go
ryan-blunden: Ok. I'll add a note in our documentation about this instead.

PR #304, opened 2022-04-15
Use existing installation location if one exists

watsonian:
This adds an additional check to see if the Doppler CLI already exists on the PATH somewhere. If it does, it uses that as the installation location. It addresses problems introduced by Homebrew switching to installing under `/opt/homebrew` and will also make it so users who installed to a custom location using `--install-path` get updated properly when `doppler update` is run.

Closes ENG-2166

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -528,6 +528,25 @@ elif [ "$format" = "tar" ]; then
       fi
     fi
 
+    # check for an existing Doppler binary
+    if [ "$binary_installed" -eq 0 ]; then
+      existing_install_dir="$(command -v doppler || true)"
+      if [ "$existing_install_dir" != "" ]; then
+        install_dir="$(dirname "$existing_install_dir")"
+        # capture exit code without exiting
+        set +e
+        install_binary "$install_dir"
+        exit_code=$?
+        set -e
+        if [ $exit_code -eq 0 ]; then
+          binary_installed=1
+          BINARY_INSTALLED_PATH="$install_dir"
+        elif [ $exit_code -eq 1 ]; then
+          found_non_writable_path=1
+        fi
+      fi
+    fi
+
     if [ "$binary_installed" -eq 0 ]; then
       install_dir="/usr/local/bin"
       # capture exit code without exiting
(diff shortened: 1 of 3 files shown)


scripts/install.sh
```diff
@@ -528,6 +528,23 @@ elif [ "$format" = "tar" ]; then
       fi
     fi
 
+    # check for an existing Doppler binary
+    existing_install_dir=$(which doppler)
```
Piccirello: `which` isn't available on some platforms, so you'll want to use `command -v` here

scripts/install.sh
```diff
@@ -528,6 +528,23 @@ elif [ "$format" = "tar" ]; then
...
+      set +e
+      install_binary "$install_dir"
+      exit_code=$?
+      set -e
+      if [ $exit_code -eq 0 ]; then
+        binary_installed=1
+        BINARY_INSTALLED_PATH="$install_dir"
+      elif [ $exit_code -eq 1 ]; then
+        found_non_writable_path=1
+      fi
```
Piccirello: Currently, if the CLI already exists but we're not able to overwrite it, we'll keep executing and ultimately install a second copy in one of the default paths. This is probably ok, but wanted to call it out.

Piccirello (commented):
Can you add a test for this so we can ensure it doesn't break in the future? I'm thinking probably a test to ensure that the new logic works as expected when the CLI is in a non-default path, and maybe another test to ensure we continue execution when the existing binary can't be updated.

scripts/install.sh
watsonian: Per our discussion we're going to leave this as-is. This situation shouldn't come up too frequently and if it does, it's not a huge issue. If it comes up as a problem later, we can revisit it.

watsonian:
I think this is good to go now. I squished the commits down and have tests. Let me know if you see anything that could use some refinement.

Piccirello (dismissed):
These tests look great. Nice job!

rgmvisser:
Good stuff @watsonian ! :)

Piccirello:
If you rebase onto `master` the Salus check should start passing

watsonian:
Nice! The rebase fixed that. Everything looks good, but it needs one last code owner review before it'll let me merge.

Piccirello (approved):

PR #308, opened 2022-05-06
Fix `run` erroneously using local environment variable

### LINKED ISSUE #307: doppler run no longer overrides local environment variables
**Describe the bug**

If you have an environment variable that matches a Doppler variable, using `doppler run` will choose the environment variable, 
not the Doppler one, even if you don't use the `--preserve-env` flag.

This is a regression introduced in 3.39.0

**To Reproduce**

1. Set an environment variable locally that has the same name as one in Doppler.
2. Use `doppler run node` and then log the environment variable
3. It will log the value from your local environment

**Expected behavior**

It should log the value from Doppler.

**CLI Version:**
3.39.0

Piccirello:
The local env var should only be given precedence when `--preserve-env` is specified. This also simplifies the key reconciliation logic a bit.

Closes #307.

### DIFF
--- a/tests/e2e/run.sh
+++ b/tests/e2e/run.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+set -euo pipefail
+
+TEST_NAME="run"
+
+cleanup() {
+  exit_code=$?
+  if [ "$exit_code" -ne 0 ]; then
+    echo "ERROR: '$TEST_NAME' tests failed during execution"
+    afterAll || echo "ERROR: Cleanup failed"
+  fi
+
+  exit "$exit_code"
+}
+trap cleanup EXIT
+
+beforeAll() {
+  echo "INFO: Executing '$TEST_NAME' tests"
+  "$DOPPLER_BINARY" run clean --max-age=0s --silent
+}
+
+beforeEach() {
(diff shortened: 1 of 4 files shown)


nmanoogian:
Ah man! I was hovering right over this yesterday but I let it go! Should have trusted my gut and tracked it down 😞

tests/e2e/run.sh
```diff
@@ -0,0 +1,58 @@
...
+
+beforeEach
+
+# verify local env is used when key isn't specified in Doppler
+value="$(NONEXISTENT_KEY=123 "$DOPPLER_BINARY" run --config prd_e2e_tests -- printenv NONEXISTENT_KEY)"
+[[ "$value" == "123" ]] || error "ERROR: local env var is not used"
+
+beforeEach
+
+# verify local env is used when key isn't specified in Doppler (and --preserve-env is specified)
```
nmanoogian, line 60: Would it be worth adding a test for one of the excluded vars (e.g. `PATH`) as well?

nmanoogian (dismissed):
Great cleanup! This logic is a lot easier to follow now

tests/e2e/run.sh
Piccirello: Great idea, added a test for this.

nmanoogian (approved):

PR #314, opened 2022-06-07
chore: update libs

Piccirello:
Upgraded with `go get -u`.


rgmvisser (approved):

nmanoogian (approved):

nmanoogian:
I recalled an issue we ran into with a newer go-keyring version so I did test this locally -- I can fetch secrets at least!

Piccirello:
> I recalled an issue we ran into with a newer go-keyring version so I did test this locally -- I can fetch secrets at least!

IIRC this was due to using an older go version to build the binary, and not an actual issue. I'm not certain if GitHub Actions makes a keyring available, bc then our E2E tests would capture this flow.

PR #317, opened 2022-07-06
Support releasing via GitHub Actions

Piccirello:
This PR moves the release process to GitHub.

We'll need to decide on the best way to test this. Based off of running this ([logs](https://github.com/DopplerHQ/cli/runs/7352792622)), the main thing we'll want to test is publishing (Docker Hub, GCR, Homebrew, scoop).


.github/workflows/release.yaml
```diff
@@ -6,14 +6,16 @@ on:
...
         options:
           - "major"
           - "minor"
           - "patch"
 
 jobs:
-  build:
+  publish:
     runs-on: ubuntu-latest
+    if: github.actor == 'piccirello'
```
Piccirello, line 18: This workflow will be runnable by anyone with 'write' perms to the repo. To lock it down further, we can use an allowlist of users. I'm starting with myself and we can expand this. Ideally we could set a group, however I haven't found a way to do this.

nmanoogian: I think this can also be defeated by creating a branch and just changing this file (to run on every branch, e.g.). This seems like a huge gap but looking around StackOverflow and GitHub's docs, I couldn't find anyone talking about it.

Still, I think it's worth it to have the automation in place

nmanoogian (approved):
Nice! It's awesome to finally have support for this!

.github/workflows/release.yaml
Piccirello: For posterity: this was solved with environments in https://github.com/DopplerHQ/cli/pull/337

PR #318, opened 2022-07-15
chore: fix scoop install on GitHub Actions

Piccirello:


nmanoogian (approved):
Not super knowledgable on Scoop but I'm assuming that the [recommended Scoop installation command](https://scoop.sh/) didn't allow for the right privs in GitHub Actions? If so, looks solid to me.

Is our new `iex` command preferable to `irm get.scoop.sh | iex -RunAsAdmin`?

Piccirello:
> Not super knowledgable on Scoop but I'm assuming that the [recommended Scoop installation command](https://scoop.sh/) didn't allow for the right privs in GitHub Actions? If so, looks solid to me.
> 
> Is our new `iex` command preferable to `irm get.scoop.sh | iex -RunAsAdmin`?

I'm not super knowledgeable on this either tbh. GitHub Actions does require running this as an admin, and this is the recommended command in [Scoop's docs](https://github.com/ScoopInstaller/Install#for-admin).

PR #321, opened 2022-07-26
Add secrets mount support for ASP.NET settings files

ryan-blunden:
`doppler run --mount` currently only supports environment variable compatible name transformers.

This PR defines a new models collection of available mount formats and adds support for the `dotnet-json` name transformer.

https://user-images.githubusercontent.com/133014/183775750-e576a9fd-f6d8-4459-9adb-9e59cfe166f9.mp4

Closes DEVRL-417

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -145,12 +145,11 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		shouldMountFile := mountPath != ""
 		shouldMountTemplate := mountTemplate != ""
 
-		var mountFormat *models.SecretsMountFormat
-		if mountFormatString != "" {
-			mountFormat = models.SecretsMountFormatMap[mountFormatString]
-			if mountFormat == nil {
-				utils.HandleError(fmt.Errorf("invalid mount format. Valid formats are %s", validMountFormatsList))
-			}
+		var mountFormat string
+		if mountFormatVal, ok := models.SecretsMountFormatMap[mountFormatString]; ok {
+			mountFormat = mountFormatVal
+		} else {
+			utils.HandleError(fmt.Errorf("Invalid mount format. Valid formats are %s", models.SecretsMountFormats))
 		}
 
 		if preserveEnv {
@@ -185,13 +184,16 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			if shouldAutoDetectFormat {
 				if shouldMountTemplate {
 					mountFormat = models.TemplateMountFormat
-					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat.Type))
+					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat))
+				} else if utils.IsDotNETSettingsFile(mountPath) {
+					mountFormat = models.DotNETJSONMountFormat
+					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat))
 				} else if strings.HasSuffix(mountPath, ".env") {
 					mountFormat = models.EnvMountFormat
-					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat.Type))
+					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat))
 				} else if strings.HasSuffix(mountPath, ".json") {
 					mountFormat = models.JSONMountFormat
-					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat.Type))
+					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat))
 				} else {
 					parts := strings.Split(mountPath, ".")
 					detectedFormat := parts[len(parts)-1]
@@ -209,7 +211,7 @@ doppler run --mount secrets.json -- cat secrets.json`,
 				utils.HandleError(errors.New("--mount-template must be specified when using --mount-format=template"))
 			}
 
-			absMountPath, handler, err := controllers.MountSecrets(secrets, mountFormat.Type, mountPath, maxReads, templateBody)
+			absMountPath, handler, err := controllers.MountSecrets(secrets, mountFormat, mountPath, maxReads, templateBody)
 			if !err.IsNil() {
 				utils.HandleError(err.Unwrap(), err.Message)
 			}
@@ -251,7 +253,7 @@ doppler run --mount secrets.json -- cat secrets.json`,
 				}
 			}
 
-			for _, envVar := range controllers.MapToEnvFormat(secrets, false) {
+			for _, envVar := range utils.MapToEnvFormat(secrets, false) {
 				env = append(env, envVar)
 			}
 		}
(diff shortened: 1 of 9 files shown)


pkg/cmd/run.go
```diff
@@ -189,6 +189,9 @@ doppler run --mount secrets.json -- cat secrets.json`,
 				} else if strings.HasSuffix(mountPath, ".env") {
 					mountFormat = models.EnvMountFormat
 					utils.LogDebug(fmt.Sprintf("Detected %s format", mountFormat.Type))
+				} else if strings.HasSuffix(mountPath, "appSettings.Development.json") || strings.HasSuffix(mountPath, "appSettings.json") {
```
nmanoogian: Would it make sense to do a regex check here?
https://go.dev/play/p/cjDrjqRTQ5Z

I mainly mention it because the .NET docs have some of their examples in lowercase or with other environments:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0

pkg/controllers/secrets.go
```diff
@@ -74,6 +74,10 @@ func MapToEnvFormat(secrets map[string]string, wrapInQuotes bool) []string {
 	return env
 }
 
+func MapToDotNETFormat(secrets map[string]string) map[string]string {
```
nmanoogian: nit/nofix: This function isn't used in this commit so you could just add it in the next one. I think it's probably fine to leave as-is though 👍

pkg/controllers/secrets.go
```diff
@@ -74,8 +74,27 @@ func MapToEnvFormat(secrets map[string]string, wrapInQuotes bool) []string {
 	return env
 }
 
-func MapToDotNETFormat(secrets map[string]string) map[string]string {
-	return secrets
+func MapDotNETJSONFormat(secrets map[string]string) map[string]string {
+	var dotnetJSON = make(map[string]string)
+	var upperSnakeToUpperCamel = func(name string) string {
```
nmanoogian: minor: Could we move this function to `util.go` and perhaps include some tests in `util_test.go`? Potentially the whole `MapDotNETJSONFormat` could go in there, what do you think?

nmanoogian (dismissed):

pkg/controllers/secrets.go
ryan-blunden: Will leave as is but will bear this in mind for next time.

pkg/cmd/run.go
ryan-blunden: Great catch! Will use regex instead.

pkg/controllers/secrets.go
ryan-blunden: @nmanoogian I like the idea of moving these to the utils package but put them in a `new name_transformers.go` file.

pkg/utils/name_transformers.go
```diff
@@ -0,0 +1,75 @@
...
+	for key, value := range secrets {
+		keyTransform := DotNETNameTransform(key)
+		dotnetJSON[keyTransform] = value
+	}
+	return dotnetJSON
+}
+
+// Test whether the file is ASP.NET Core settings file
+func IsDotNETSettingsFile(fileName string) bool {
+	matched, _ := regexp.MatchString(`(?i)appsettings(\.[a-z0-9-]+)?\.json`, fileName)
```
nmanoogian: Ah, I forgot underscores in that regex, if you think people will want that: `(?i)appsettings(\.[a-z0-9_-]+)?\.json`

pkg/utils/name_transformers_test.go
```diff
@@ -0,0 +1,88 @@
...
+limitations under the License.
+*/
+package utils
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestDotNETNameTransform(t *testing.T) {
```
nmanoogian: These tests are awesome! 🎉

nmanoogian (dismissed):
Sweet change! I forgot to mention: Any commits that don't start with `chore:` will end up in the release notes automatically. I'm thinking just your first commit should be included in the notes

pkg/cmd/run.go
```diff
@@ -137,14 +137,22 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		dopplerSecrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL)
 
 		mountPath := cmd.Flag("mount").Value.String()
-		mountFormat := cmd.Flag("mount-format").Value.String()
+		mountFormatString := cmd.Flag("mount-format").Value.String()
```
Piccirello, line 140: Can you separate the `Added secrets mount support for ASP.NET core JSON settings files` commit into two commits? One commit will be a `chore` to refactor the existing strings into constants, and the other will be to add support for the new format. That will make reviewing the important stuff (i.e. the new logic) really easy.

Separately, you'll want to use the imperative mood in your commits[^0]. So `Added` -> `Add`

[^0]: https://cbea.ms/git-commit/#imperative

pkg/utils/name_transformers.go
```diff
@@ -70,6 +70,6 @@ func MapDotNETJSONFormat(secrets map[string]string) map[string]string {
 
 // Test whether the file is ASP.NET Core settings file
 func IsDotNETSettingsFile(fileName string) bool {
-	matched, _ := regexp.MatchString(`(?i)appsettings(\.[a-z0-9-]+)?\.json`, fileName)
+	matched, _ := regexp.MatchString(`(?i)appsettings(\.[a-z0-9_-]+)?\.json`, fileName)
```
Piccirello: You'll want to squash this commit into the previous commit. No need for a new commit if the code you're changing isn't yet in `main` :)

pkg/cmd/run.go
ryan-blunden: Done!

I also moved the `IsDotNETSettingsFile` function into its own `pkg/utils/secrets_mount.go` file as it it's related to name transformers.

pkg/cmd/run.go
```diff
@@ -137,14 +137,22 @@ doppler run --mount secrets.json -- cat secrets.json`,
...
 		// only auto-detect the format if it hasn't been explicitly specified
 		shouldAutoDetectFormat := !cmd.Flags().Changed("mount-format")
 		shouldMountFile := mountPath != ""
 		shouldMountTemplate := mountTemplate != ""
 
+		var mountFormat *models.SecretsMountFormat
+		if mountFormatString != "" {
+			mountFormat = models.SecretsMountFormatMap[mountFormatString]
+			if mountFormat == nil {
+				utils.HandleError(fmt.Errorf("invalid mount format. Valid formats are %s", validMountFormatsList))
```
Piccirello: In `MountSecrets()` we use a slightly different error message. Can we use the same error in both places? I don't have a preference for which version we use, though you'll want to capitalize `invalid` to `Invalid`.

In `MountSecrets()` we use:
```
fmt.Errorf("Invalid mount format: %s", format)
```

pkg/models/secrets_mount.go
```diff
@@ -0,0 +1,55 @@
...
+var DotNETJSONMountFormat = &SecretsMountFormat{
+	Name: "dotnet-json",
+	Type: "dotnet-json",
+}
+var SecretsMountFormatList = []*SecretsMountFormat{
+	EnvMountFormat,
+	JSONMountFormat,
+	DotNETJSONMountFormat,
+	TemplateMountFormat,
+}
```
Piccirello: Can these all be declared `const`?

Piccirello (commented):
Last two comments from me!

pkg/cmd/run.go
ryan-blunden: Done!

pkg/models/secrets_mount.go
ryan-blunden: It doesn't look like Go supports const for structs so I've refactored this to use strings instead which seems to simplify things without penalty.

Piccirello (approved):

Issue #326, opened 2022-08-22
[FEATURE] Hide secrets from logs when running command with `doppler run --hide-secrets `


remorses:
**Is your feature request related to a problem? Please describe.**
When using doppler on Github Actions with a command like `doppler run -- printenv` your secrets will be leaked in the Github actions logs

Usually Github Actions replaces Github Actions secrets with `****` in the logs, but if you use doppler you won't have this feature

This is also useful if you are streaming on Twitch and want to show your terminal logs safely

**Describe the solution you'd like**
When running `doppler run --hide-secrets` replace all doppler secrets instances with `****` characters

**Describe alternatives you've considered**

**Additional context**
It would also be useful to add the option `--show SECRET_NAME` to allow some secrets to be shown, useful if you are storing non sensitive data in doppler and want to show them in logs

Issue #322, opened 2022-07-28
[FEATURE] doppler run --secret-names


icepuma:
**Is your feature request related to a problem? Please describe.**
Let's say I have a project / config containing 10 secrets, but I want just 2 of them in my `doppler run` - call, the command will receive all 10 secrets. But sometimes I don't want to give the command all my secrets.

**Describe the solution you'd like**
The solution might be an additional switch / filter e.g. `--secret-names` where I can provide a list (1-n) secret names which should be injected into the `doppler run` - call.

**Describe alternatives you've considered**
> Nothing to say here

**Additional context**
> Nothing to say here

icepuma:
In the case you're cool with this filter feature, I'm keen to provide the feature via PR.

ryan-blunden:
Hiey @icepuma and thanks for the feature request!

Are you able to expand upon why you only want to provide a subset of secrets?

We do have long-term plans for building in the reading of specific secrets into our API and then CLI but this isn't yet on our engineering roadmap.

For now, I'd recommend creating a branch config containing only the subset of secrets required.

icepuma:
It is more for ad-hoc actions when running `doppler run` on my local machine, where I have a config containing 10 values, but only want to inject 2 of those values to a command.

Creating a branch config would mean I have to do something just for this one ad-hoc action and maybe to forget to delete it afterwards.

ryan-blunden:
Is it that the inclusion of the 8 secrets breaks the command you need to run? Just wanting to understand the need for removing those secrets.

icepuma:
I wouldn't break the command, it is more that I don't need the 8 other secrets in this one particular run and therefore don't want the command let them know. It is more like a filter, where you have a more granular control over what leaves your doppler ecosystem and what doesn't.

Or maybe it would be better add a `--better-safe-than-sorry` switch where you show the user the set of secrets, that will be injected into the env of the command.

I just want to have a way where people don't yolo their secret injection. The filtering would've helped me personally, but maybe showing the user which secrets will be injected creates more awareness over the usage of doppler.

ryan-blunden:
I see where you're coming from.

You could use `doppler secrets --only-names` to show the list of secrets supplied to a command.

You could also use `doppler secrets get` instead of `doppler run` to provide a reduced list of secrets:

```sh
API_KEY="$(doppler secrets get API_KEY--plain)" AUTH_TOKEN="$(doppler secrets get AUTH_TOKEN--plain)" ./bin/script.sh
```

PR #327, opened 2022-08-22
Implement `hide-secrets` flag for run command

remorses:
Implements #326

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -19,9 +19,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


remorses:
I can add tests for `ReplaceTransformer` if you want, i copied that code from [here](https://github.com/icholy/replace/blob/a7e12fe69d82503d82c3f85a9ca3973a11a2085f/replace.go#L12)

Piccirello:
I'm having difficulty understanding the practical use case for this, beyond demos. Why would a user want invalid values to be injected? I think a more pragmatic solution would be to support specifying an allowlist of secrets that should be injected (#322).

remorses:
This pr does not inject invalid values, it simply hides them from the terminal output, the env variables remain the same

some use cases
- Making sure CI logs do not contain secrets (anyone that can see your logs could steal your secrets, very dangerous)
- streaming on twitch or demos

benhylak:
this is nice.

PR #332, opened 2022-09-13
Use standard zsh path for rpm/RHEL releases

### LINKED ISSUE #331: [BUG/FEATURE] Add support for OSTree based environments
**Describe the bug**
Doppler CLI does not install due to it installing files into a path classified as a system directory in OSTree based OSes. See: https://github.com/coreos/rpm-ostree/issues/233
This is caused by installing persistent files into the /usr/local/ directory, which is classified as a system directory, and therefore not writable on OSTree based OSes.

**To Reproduce**
Steps to reproduce the behavior. Please include output from running the command with `--debug`.
1. Add the Doppler Repo to any OSTree based OS, such as Fedora CoreOS.
2. Attempt to install Doppler with rpm-ostree

**Expected behavior**
Doppler should install properly without triggering the bug/unsupported behavior as described in https://github.com/coreos/rpm-ostree/issues/233

**Screenshots**
image

 - Guest OS: Fedora CoreOS
 - Version: 36.20220820.3.0
 - Host OS: Proxmox
 - Version: 7.2-7

**CLI Version:**
Version: ALL
(As this is an issue with installing it, I would assume that this would effect most versions, but I am currently installing the most recent version for Fedora/RHEL/CentOS)

**Additional context**

mariaWitch:
Due to a quirk with OSTree releases, the rpm releases should place the doppler zsh completion file in /usr/share/zsh/, which is the default for Fedora/RHEL/CentOS based systems. In theory this should fix #331 but I will need to perform testing to confirm. There may be a more elegant way to accomplish this, but this seems to be the most straight forward way that I can find for the moment.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -166,6 +166,13 @@ nfpms:
         dst: /etc/bash_completion.d/doppler
       - src: ./completions/doppler.zsh
         dst: /usr/local/share/zsh/site-functions/_doppler
+        packager: deb
+      - src: ./completions/doppler.zsh
+        dst: /usr/local/share/zsh/site-functions/_doppler
+        packager: apk
+      - src: ./completions/doppler.zsh
+        dst: /usr/share/zsh/site-functions/_doppler
+        packager: rpm
       - src: ./completions/doppler.fish
         dst: /usr/share/fish/completions/doppler.fish
 


mariaWitch:
Validated on Fedora CoreOS. zsh also autocompletes, so functionality is fully working.

mariaWitch:
Validated on AlmaLinux. Going to do Fedora and then CentOS next. (These are non-OSTree based distros).

Update: Confirmed working on zsh on Fedora and CentOS Stream as well.

mariaWitch:
@Piccirello Looking for review on this/the related issue.

Piccirello:
Thanks for putting up this fix, and for testing across those different OSes! Could you squash the two merge commits so that this PR only has one commit? We can then get this merged in and released.

mariaWitch:
@Piccirello Done.

Piccirello (dismissed):
LGTM

mariaWitch:
Is there is a specific reason for why the e2e test keeps failing? And if that will block merging, how do I resolve the issue?

mariaWitch:
I believe I have solved the issue for why the e2e test was failing (it wasn't related to any of the changes, and was environment specific). If you could manually re-run the test on your end, it should pass now. @Piccirello

mariaWitch:
I was able to get it to work on the forked repository with the correct secret values. So I don't understand why it isn't working here. It seems to be a github environment related issue though.
image

mariaWitch:
Same as before, just updated again.

Piccirello (approved):
I believe the tests are failing because GitHub isn't allowing you to access the required secrets to run the tests. This change won't impact our tests so will get this merged.

Piccirello:
This change has been released in CLI v3.45.0.

PR #337, opened 2022-10-04
chore: run release workflow from production environment

Piccirello:
This allows us to better control which branches can access the synced secrets, as well as requiring an approval flow.


nmanoogian (approved):
I suppose Environments does solve the "anyone can write a workflow" issue -- nice!

Piccirello:
> I suppose Environments does solve the "anyone can write a workflow" issue -- nice!

It does appear to, and also allows us to setup an explicit approval process before executing the workflow. This workaround should be better documented!

PR #338, opened 2022-10-07
Add handling for gnupg error

watsonian:
This ensures the error that's encountered when updating via the install.sh script if gnupg isn't installed is passed through to the CLI binary.

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -73,8 +73,11 @@ func RunInstallScript() (bool, string, Error) {
 
 		message := "Unable to install the latest Doppler CLI"
 		permissionError := exitCode == 2 || strings.Contains(strOut, "dpkg: error: requested operation requires superuser privilege")
+		gnupgError := exitCode == 3
 		if permissionError {
 			message = "Error: update failed due to improper permissions\nPlease re-run with `sudo` or as an admin"
+		} else if gnupgError {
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -391,7 +391,7 @@ if [ -x "$gpg_binary" ]; then
 else
   log "ERROR: Unable to find gpg binary for signature verification"
   log "You can resolve this error by installing your system's gnupg package"
-  clean_exit 1
+  clean_exit 3
```
Piccirello, line 395: Can you add this error code to the comment on lines 5-7?

pkg/controllers/update.go
```diff
@@ -77,6 +77,11 @@ func RunInstallScript() (bool, string, Error) {
 			message = "Error: update failed due to improper permissions\nPlease re-run with `sudo` or as an admin"
 		}
 
+		gnupgError := exitCode == 3
+		if gnupgError {
+			message = "Error: Unable to find gpg binary for signature verification\nYou can resolve this error by installing your system's gnupg package"
+		}
```
Piccirello: Can you make this an `else if` off of the above conditional? Just so we're certain that only one branch runs. Something like:

```
permissionError := exitCode == 2 || ...
gnupgError := exitCode == 3
if permissionError {
  ...
} else if gnupgError {
 ...
}
```

Piccirello (approved):

nmanoogian (approved):

PR #340, opened 2022-10-25
Increase default number of pbkdf2 rounds to 500,000

Piccirello:


pkg/crypto/aes.go
```diff
@@ -27,17 +27,21 @@ import (
...
 )
 
-const Base64EncodingPrefix = "base64:"
-const HexEncodingPrefix = "hex:"
+const Base64EncodingPrefix = "base64"
+const HexEncodingPrefix = "hex"
 
-func deriveKey(passphrase string, salt []byte) ([]byte, []byte, error) {
+const Pbkdf2Rounds = 500000
+const LegacyPbkdf2Rounds = 50000
```
Kunamatata: These are exported but not used anywhere else.
Should they be lowercase?

Kunamatata:
Would it be valuable to refactor `aes.go` into leveraging structs and methods? This could potentially have the nice effect of encapsulating and separating encoding and decoding concerns from encryption and decryption and provide an interface for those.

We could eventually go one step further and extract the encoding and decoding (base64, hex) into their own structs and pass and provide an `AESEncoderDecoder` interface and then pass that to the `AESEncryptDecrypt` struct.

The call sites would have to change to accommodate initializing the new struct.

Here's a loose initial example.
```go
type AESEncrytertDecrypter interface {
	Encrypt(passphrase string, plaintext []byte, encoding string) (string, error)
	Decrypt(passphrase string, ciphertext []byte) (string, error)
}

type AESEncryptDecrypt struct {
	base64EncodingPrefix string
	hexEncodingPrefix    string
	pbkdf2Rounds         int
	legacyPbkdf2Rounds   int
}

func NewAESEncrypterDecypter() AESEncrytertDecrypter {
	return &AESEncryptDecrypt{
		base64EncodingPrefix: "base64",
		hexEncodingPrefix:    "hex",
		pbkdf2Rounds:         500000,
		legacyPbkdf2Rounds:   50000,
	}
}

```

pkg/crypto/aes.go
```diff
@@ -81,35 +89,27 @@ func Encrypt(passphrase string, plaintext []byte, encoding string) (string, erro
...
 		encodedIV = base64.StdEncoding.EncodeToString(iv)
 		encodedData = base64.StdEncoding.EncodeToString(data)
 	} else if encoding == "hex" {
-		prefix = HexEncodingPrefix
 		encodedSalt = hex.EncodeToString(salt)
 		encodedIV = hex.EncodeToString(iv)
 		encodedData = hex.EncodeToString(data)
 	} else {
 		return "", errors.New("Invalid encoding, must be one of [base64, hex]")
 	}
```
Kunamatata, line 105: It seems to be more idiomatic and recommended to use switch instead of if else if else statements:
> Switch
> Go's switch is more general than C's. The expressions need not be constants or even integers, the cases are evaluated top to bottom until a match is found, and if the switch has no expression it switches on true. It's therefore possible—and idiomatic—to write an if-else-if-else chain as a switch.

nit: Convert if else if else chain to tagged switch:
https://staticcheck.io/docs/checks#QF1003 (linter)

```go
switch encoding {
case "base64":
	encodedSalt = base64.StdEncoding.EncodeToString(salt)
	encodedIV = base64.StdEncoding.EncodeToString(iv)
	encodedData = base64.StdEncoding.EncodeToString(data)
case "hex":
	encodedSalt = hex.EncodeToString(salt)
	encodedIV = hex.EncodeToString(iv)
	encodedData = hex.EncodeToString(data)
default:
	return "", errors.New("Invalid encoding, must be one of [base64, hex]")
}
```

Kunamatata (changes requested):

pkg/crypto/aes_test.go
```diff
@@ -24,30 +24,46 @@ const originalPassphrase = "secret"
...
+	plaintext, err = Decrypt(originalPassphrase, []byte(ciphertext))
 	if err != nil || plaintext != originalPlaintext {
 		t.Error("Invalid plaintext when decrypting hex value")
 	}
 
 	// decode base64 w/o prefix (should error)
-	ciphertext = "qwbkFMWB7FE=-Ew968YdkAXRb6l46-eA4o9Pf9mSIaOofa8YIEP+FqJ6DwScHsYIObAw3dvKvHbe5SDTzB"
-	plaintext, err = Decrypt(originalPassphrase, []byte(ciphertext), "base64")
+	ciphertextData = "qwbkFMWB7FE=-Ew968YdkAXRb6l46-eA4o9Pf9mSIaOofa8YIEP+FqJ6DwScHsYIObAw3dvKvHbe5SDTzB"
+	plaintext, err = Decrypt(originalPassphrase, []byte(ciphertextData))
```
Kunamatata: nit: `plainText` is not used -> `_, err = ...`

Kunamatata:
optional nit: 
- You didn't touch this portion but `time.Now().Sub(now).Milliseconds()` can be replaced with `time.Since(now).Milliseconds()` https://github.com/DopplerHQ/cli/blob/f98446169a5cdcc9ff2c6c4689d01b3b14adb6c2/pkg/crypto/aes.go#L70

pkg/crypto/aes.go
Kunamatata: Same thing for this portion: https://github.com/DopplerHQ/cli/blob/f98446169a5cdcc9ff2c6c4689d01b3b14adb6c2/pkg/crypto/aes.go#L195-L204

```go
switch rawEncoding {
case Base64EncodingPrefix:
	encoding = Base64EncodingPrefix
case hexEncodingPrefix, "":
	// default to hex for backwards compatibility b/c we didn't always include an encoding prefix
	// TODO remove support for optional prefix when releasing CLI v4 (DPLR-435)
	encoding = hexEncodingPrefix
default:
	return "", errors.New("Invalid encoding, must be one of [base64, hex]")
}
```

pkg/crypto/aes.go
```diff
@@ -172,28 +158,53 @@ func decodeHex(passphrase string, ciphertext string) ([]byte, []byte, []byte, er
...
+// Formats:
+// 1) `encoding:text`
+// 2) `text`
+func Decrypt(passphrase string, ciphertext []byte) (string, error) {
 	var salt []byte
 	var iv []byte
 	var data []byte
-	if encoding == "base64" {
+
+	cParts := strings.SplitN(string(ciphertext), ":", 3)
```
nmanoogian, line 177: ~nit: Should this be 2?~

Update: I see why this is 3 now, even though 2 were only required in the first commit 👍

pkg/crypto/aes.go
```diff
@@ -172,28 +158,53 @@ func decodeHex(passphrase string, ciphertext string) ([]byte, []byte, []byte, er
...
+	rawEncoding := ""
+	var ciphertextData string
```
nmanoogian: supernit: Any reason why you initialized these vars differently? They will all be initialized as empty strings as you have them now

nmanoogian (dismissed):
Nice! Loved the commit history here -- very easy to reason about

pkg/crypto/aes.go
Piccirello: This seems more like personal preference rather than correctness. I'm going to stick with the current approach rather than refactor all of our conditionals since it doesn't seem we gain anything. If we want to make this codebase more idiomatic, I'd push for investing our time in adding support for a `context`.

Kunamatata (dismissed):

pkg/crypto/aes.go
Piccirello: None whatsoever. Will update this.

nmanoogian: Cool 👍 Just wanted to make sure there wasn't a misunderstanding here

nmanoogian (approved):

PR #342, opened 2022-10-25
ENG-5298: Fix deprecations and clean up cli code

Kunamatata:
## Description

This PR addresses warnings and suggestions made by `staticcheck` as well as `go vet`
- Address warnings and suggestions from the result of: `staticcheck ./... ; go vet ./...` 
- Replace `ioutil` package with `io` and `os` by following this blog post: https://go.dev/doc/go1.16#ioutil
- Replace `golang.org/x/crypto/ssh/terminal` with `golang.org/x/term` instead
  **Deprecation notice**: https://pkg.go.dev/golang.org/x/crypto/ssh/terminal#pkg-overview

## Suggestion / Recommendation
We should incorporate linters such as `staticcheck` or `golangci-lint` (which is an aggregation of linters which contains staticcheck) and `go vet` to our github CI actions

We should address the `golangci-lint` results below as well but I want to start this conversation instead of fixing those right now.

## golangci-lint example
Here is an example of `golangci-lint run ./...`. 
⚠️ **These are not currently covered by this PR**
```
❯ golangci-lint run ./...                                                                                                                                                       15:29:28
pkg/printer/print.go:121:4: S1038: should use fmt.Printf instead of fmt.Println(fmt.Sprintf(...)) (but don't forget the newline) (gosimple)
                        fmt.Println(fmt.Sprintf("· %s", change))
                        ^
pkg/cmd/secrets.go:37:6: type `secretsResponse` is unused (unused)
type secretsResponse struct {
     ^
pkg/cmd/secrets.go:469:3: ineffectual assignment to enableFallback (ineffassign)
                enableFallback = false
                ^
pkg/cmd/secrets.go:470:3: ineffectual assignment to enableCache (ineffassign)
                enableCache = false
                ^
pkg/utils/random.go:29:11: Error return value of `rand.Read` is not checked (errcheck)
        rand.Read(buffer) // #nosec G104
                 ^
pkg/utils/util.go:165:23: Error return value of `cmd.Process.Signal` is not checked (errcheck)
                                cmd.Process.Signal(sig) // #nosec G104
                                                  ^
pkg/utils/util.go:175:21: Error return value of `cmd.Process.Signal` is not checked (errcheck)
                cmd.Process.Signal(os.Kill) // #nosec G104
                                  ^
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,9 @@ test-packages:
 	./tests/packages.sh
 
 test-release:
-	doppler run -- goreleaser release --snapshot --skip-publish --rm-dist --parallelism=4
+	@if [ ! -f "$$GOPATH/bin/goreleaser" ]; then echo "Error: goreleaser is not installed\n\nYou can install goreleaser with 'go install github.com/goreleaser/goreleaser@latest'" && exit 1; fi
+	$$GOPATH/bin/doppler run -- goreleaser release --snapshot --skip-publish --rm-dist --parallelism=4
 
 scan:
-	if [ ! -f "$$GOPATH/bin/gosec" ]; then echo "Error: gosec is not installed\n\nYou can install gosec with 'go get github.com/securego/gosec/cmd/gosec'\n" && exit 1; fi
+	@if [ ! -f "$$GOPATH/bin/gosec" ]; then echo "Error: gosec is not installed\n\nYou can install gosec with 'go install github.com/securego/gosec/cmd/gosec@latest'\n" && exit 1; fi
 	$$GOPATH/bin/gosec -quiet ./pkg/...
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -302,7 +301,7 @@ var runCleanCmd = &cobra.Command{
 			utils.HandleError(err, "Unable to read fallback directory")
 		}
 
-		entries, err := ioutil.ReadDir(defaultFallbackDir)
+		entries, err := os.ReadDir(defaultFallbackDir)
 		if err != nil {
 			utils.HandleError(err, "Unable to read fallback directory")
 		}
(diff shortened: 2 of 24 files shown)


pkg/cmd/run.go
```diff
@@ -223,7 +223,7 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			// remove any reserved keys from secrets
 			reservedKeys := []string{"PATH", "PS1", "HOME"}
 			for _, reservedKey := range reservedKeys {
-				if _, found := dopplerSecrets[reservedKey]; found == true {
```
nmanoogian, line 227: nit: typo in commit message: `chore: Omit comparaison with boolean constant`

Makefile
```diff
@@ -13,8 +13,9 @@ test-packages:
 	./tests/packages.sh
 
 test-release:
+	@if [ ! -f "$$GOPATH/bin/goreleaser" ]; then echo "Error: goreleaser is not installed\n\nYou can install goreleaser with 'go install github.com/goreleaser/goreleaser@latest'" && exit 1; fi
 	doppler run -- goreleaser release --snapshot --skip-publish --rm-dist --parallelism=4
 
 scan:
-	if [ ! -f "$$GOPATH/bin/gosec" ]; then echo "Error: gosec is not installed\n\nYou can install gosec with 'go get github.com/securego/gosec/cmd/gosec'\n" && exit 1; fi
-	$$GOPATH/bin/gosec -quiet ./pkg/...
```
nmanoogian, line 20: Why remove the `GOPATH` part?

nmanoogian (dismissed):
These changes look solid to me! I'll defer to @Piccirello for final approval

pkg/controllers/update.go
```diff
@@ -51,6 +51,9 @@ func RunInstallScript() (bool, string, Error) {
 
 	// write script to temp file
 	tmpFile, err := utils.WriteTempFile("install.sh", script, 0555)
+	if err != nil {
```
nmanoogian, line 54: commits: `Check utils.WriteTempFile error` commit should be prefixed with `chore:`

Makefile
Kunamatata: Good catch @nmanoogian! I updated the go install commands and was testing the commands locally and they didn't work so I had assumed something was up with the `GOPATH` but that was my local setup, so these changes are not adequate.

Re-adding the `$$GOPATH/bin` part

pkg/cmd/run.go
Kunamatata: Thanks!

nmanoogian (approved):

PR #343, opened 2022-10-26
Don't create default config dir if --configuration used

watsonian:
This solves an issue where the CLI would always attempt to create the default config directory (`$HOME/.doppler`) even if a different location was specified using `--configuration`. Now, we provide a `--config-dir` flag that will set the entire directory being used. It also deprecates the `--configuration` flag, which was around from before we used a config directory (there is no real reason to have more than one config FILE).

### DIFF
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -32,9 +32,6 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-// baseConfigDir (e.g. /home/user/)
-var baseConfigDir string
-
 // UserConfigDir (e.g. /home/user/.doppler)
 var UserConfigDir string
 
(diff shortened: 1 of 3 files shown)


pkg/configuration/config.go
```diff
@@ -59,7 +59,12 @@ func init() {
 }
 
 // Setup the config directory and config file
-func Setup() {
+func Setup(cmd *cobra.Command) {
+	if cmd.Flags().Changed("config-dir") {
+		UserConfigFile = filepath.Join(UserConfigDir, configFileName)
+	}
```
Piccirello: Is this change needed if you're setting the `UserConfigDir` and `UserConfigFile` in root's `loadFlags()` function?

watsonian: I'm not familiar enough with Cobra to know if it _should_ be required, but I know in practice it seems to be. With that change backed out:

```
❯ ./doppler secrets -p example -c dev --config-dir /tmp/doppler --debug
Debug: Using config dir /tmp/doppler
Debug: Using config file /Users/watsonian/.doppler/.doppler.yaml
```

with the change in place:

```
❯ ./doppler secrets -p example -c dev --config-dir /tmp/doppler --debug
Debug: Using config dir /tmp/doppler
Debug: Using config file /tmp/doppler/.doppler.yaml
```

Piccirello: I pushed up a [change](https://github.com/DopplerHQ/cli/compare/9ba954ce6859fdcc34e64a7a0060047273a7e403..4ba5985d517e2d11dffd3561e089798b2d12d035) to simplify this a little bit. I also added some tests to verify that `--config-dir` and `--configuration` play nicely.

Piccirello (approved):

nmanoogian (approved):

PR #344, opened 2022-11-04
Support reading network proxy from environment

Piccirello:
This has the potential to affect existing CLI usage, but seems unlikely in practice.


nmanoogian (approved):

pkg/http/http.go
```diff
@@ -194,6 +194,7 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 		DisableKeepAlives: true,
 		TLSClientConfig:   tlsConfig,
 		DialContext:       dialContext,
+		Proxy:             http.ProxyFromEnvironment,
```
Piccirello, line 197: Note from @watsonian: we should add a debug log when using a proxy

Piccirello: #345

PR #345, opened 2022-11-05
chore: print proxy info

Piccirello:
Follow up to #344.

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -188,13 +188,23 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 		return dialer.DialContext(ctx, network, addr)
 	}
 
+	proxyUrl, err := http.ProxyFromEnvironment(req)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


watsonian (dismissed):

pkg/http/http.go
```diff
@@ -188,13 +188,18 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 		return dialer.DialContext(ctx, network, addr)
 	}
 
+	proxyUrl, err := http.ProxyFromEnvironment(req)
+	if err == nil && proxyUrl != nil {
```
nmanoogian: If `err != nil`, should we log it? As it stands, in that scenario, we'll pass `nil` into `http.ProxyURL` which results in no proxy being used and the error will effectively be silenced

nmanoogian (dismissed):

pkg/http/http.go
Piccirello: Will add a debug log for this

nmanoogian (approved):

PR #346, opened 2022-11-07
chore: add explicit version number to encrypted files

Piccirello:
The format has changed as we've introduced new fields, so the explicit version number should help with parsing.

### DIFF
--- a/pkg/controllers/secrets.go
+++ b/pkg/controllers/secrets.go
@@ -81,35 +81,43 @@ func GetSecretNames(config models.ScopedOptions) ([]string, Error) {
 	return secretsNames, Error{}
 }
 
-func MountSecrets(secrets map[string]string, format string, mountPath string, maxReads int, templateBody string) (string, func(), Error) {
+// SecretsToBytes converts secrets to byte array
+func SecretsToBytes(secrets map[string]string, format string, templateBody string) ([]byte, Error) {
+	if format == models.TemplateMountFormat {
+		return []byte(RenderSecretsTemplate(templateBody, secrets)), Error{}
+	}
+
+	if format == models.EnvMountFormat {
+		return []byte(strings.Join(utils.MapToEnvFormat(secrets, true), "\n")), Error{}
+	}
+
+	if format == models.JSONMountFormat {
+		envStr, err := json.Marshal(secrets)
+		if err != nil {
+			return nil, Error{Err: err, Message: "Unable to marshall secrets to json"}
+		}
+		return envStr, Error{}
+	}
+
+	if format == models.DotNETJSONMountFormat {
+		envStr, err := json.Marshal(utils.MapToDotNETJSONFormat(secrets))
+		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/models/encrypted_file.go
```diff
@@ -0,0 +1,145 @@
...
+	return -1, errors.New("Invalid ciphertext")
+}
+
+func (options *FileVersionOptions) Parse(ciphertext string) (EncryptedFile, error) {
+	var version string
+	var encoding string
+	var numRounds string
+	var data string
+
+	cParts := strings.SplitN(string(ciphertext), ":", 4)
```
nmanoogian: Curious: Why make a copy of `ciphertext` with `string`?

pkg/crypto/aes.go
```diff
@@ -166,70 +155,50 @@ func decodeHex(passphrase string, ciphertext string) ([]byte, []byte, []byte, er
 
 // Decrypt ciphertext with a passphrase.
 // Formats:
-// 1) `encoding:numRounds:text`
-// 2) `encoding:text`
-// 3) `text`
+// 1) `encoding:numRounds:text` (latest)
```
nmanoogian: Should this include `version:encoding:numRounds:text`?

pkg/controllers/secrets.go
```diff
@@ -81,35 +81,43 @@ func GetSecretNames(config models.ScopedOptions) ([]string, Error) {
 	return secretsNames, Error{}
 }
 
-func MountSecrets(secrets map[string]string, format string, mountPath string, maxReads int, templateBody string) (string, func(), Error) {
-	if !utils.SupportsNamedPipes {
-		return "", nil, Error{Err: errors.New("This OS does not support mounting a secrets file")}
+// SecretsToBytes converts secrets to byte array
+func SecretsToBytes(secrets map[string]string, format string, templateBody string) ([]byte, Error) {
```
nmanoogian, line 85: This might be a little more than you want to bite off for this PR but I wanted to call out that Golang's FAQs recommend returning the `error` type rather than a concrete type (like `Error`). This avoids some [particularly annoying gotchas](https://go.dev/doc/faq#nil_error) and would also allow us to return `nil` from this function (rather than `Error{}`).

It looks like we might have implemented `IsNil` to work around this issue.

nmanoogian (dismissed):
Nice change! It's much easier to reason about the formats now that they're defined in code

pkg/models/encrypted_file.go
Piccirello: In a previous iteration `ciphertext` was a byte array, but this is no longer needed

pkg/controllers/secrets.go
Piccirello: This is a larger refactor so going to hold off on this for now. We use the Error pattern quite a bit throughout this codebase.

nmanoogian (approved):

PR #347, opened 2022-11-09
Drop use of legacy endpoints

Piccirello:
And add paging support for our logs commands.

### DIFF
--- a/pkg/controllers/activity.go
+++ b/pkg/controllers/activity.go
@@ -24,7 +24,7 @@ import (
 func GetActivityLogIDs(config models.ScopedOptions) ([]string, Error) {
 	utils.RequireValue("token", config.Token.Value)
 
-	logs, err := http.GetActivityLogs(config.APIHost.Value, utils.GetBool(config.VerifyTLS.Value, true), config.Token.Value)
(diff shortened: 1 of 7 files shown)


pkg/controllers/activity.go
```diff
@@ -24,7 +24,7 @@ import (
 func GetActivityLogIDs(config models.ScopedOptions) ([]string, Error) {
 	utils.RequireValue("token", config.Token.Value)
 
-	logs, err := http.GetActivityLogs(config.APIHost.Value, utils.GetBool(config.VerifyTLS.Value, true), config.Token.Value, 1)
+	logs, err := http.GetActivityLogs(config.APIHost.Value, utils.GetBool(config.VerifyTLS.Value, true), config.Token.Value, 1, 50)
```
nmanoogian: Any reason why we're providing 50 when the command defaults are 20?

Separately, would it make sense to pass 0 for page and number to let the server default these?

nmanoogian (dismissed):

pkg/controllers/activity.go
Piccirello: I provided 50 so that there'd be more relevant results, since the user doesn't have the ability to request a second page here. I like the idea of letting the server handle the defaults though - will switch to 0.

nmanoogian (approved):

PR #349, opened 2022-12-15
Don't request secret values when only names are needed

Piccirello:
This PR also limits commands to only revealing the necessary secrets, rather than revealing all and then filtering client-side.

### DIFF
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -164,18 +164,23 @@ func secrets(cmd *cobra.Command, args []string) {
 
 	utils.RequireValue("token", localConfig.Token.Value)
 
-	response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, nil, false, 0)
-	if !err.IsNil() {
-		utils.HandleError(err.Unwrap(), err.Message)
-	}
-	secrets, parseErr := models.ParseSecrets(response)
-	if parseErr != nil {
-		utils.HandleError(parseErr, "Unable to parse API response")
-	}
-
 	if onlyNames {
-		printer.SecretsNames(secrets, jsonFlag)
+		secretNames, err := http.GetSecretNames(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, false)
+		if !err.IsNil() {
+			utils.HandleError(err.Unwrap(), err.Message)
+		}
+
+		printer.SecretsNames(secretNames, jsonFlag)
 	} else {
+		response, err := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, nil, false, 0)
+		if !err.IsNil() {
+			utils.HandleError(err.Unwrap(), err.Message)
+		}
+		secrets, parseErr := models.ParseSecrets(response)
+		if parseErr != nil {
+			utils.HandleError(parseErr, "Unable to parse API response")
+		}
+
 		printer.Secrets(secrets, []string{}, jsonFlag, false, raw, false)
 	}
 }
--- a/pkg/http/api.go
+++ b/pkg/http/api.go
@@ -171,16 +171,11 @@ func GetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
 	var params []queryParam
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "config", Value: config})
+	params = append(params, queryParam{Key: "include_dynamic_secrets", Value: strconv.FormatBool(includeDynamicSecrets)})
 
 	if secrets != nil {
 		params = append(params, queryParam{Key: "secrets", Value: strings.Join(secrets, ",")})
 	}
-
-	includeDynamicSecretsOption := "false"
-	if includeDynamicSecrets {
-		includeDynamicSecretsOption = "true"
-	}
-	params = append(params, queryParam{Key: "include_dynamic_secrets", Value: includeDynamicSecretsOption})
 
 	if dynamicSecretsTTL > 0 {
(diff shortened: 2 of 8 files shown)


pkg/cmd/secrets.go
```diff
@@ -538,8 +538,29 @@ func substituteSecrets(cmd *cobra.Command, args []string) {
...
+	if !fetchAllSecrets {
+		// parse secrets from template
+		for _, field := range controllers.TemplateFields(template) {
+			if strings.HasPrefix(field, "{{.") && strings.HasSuffix(field, "}}") {
+				f := strings.TrimPrefix(field, "{{.")
+				f = strings.TrimSuffix(f, "}}")
+				secretNames = append(secretNames, f)
+			}
+		}
+	}
```
Piccirello: Unfortunately go doesn't provide a better way of grabbing the parsed fields. Thus we provide the `fetch-all` flag (open to alternative naming) to override our new parsing logic and instead fetch all secrets. This allows us to continue to support more advanced template logic than what we can manually parse.

pkg/http/api.go
```diff
@@ -233,6 +233,41 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+		return nil, Error{Err: err, Message: "Unable to fetch secret names", Code: statusCode}
+	}
+
+	var result map[string]interface{}
+	err = json.Unmarshal(response, &result)
+	if err != nil {
+		return nil, Error{Err: err, Message: "Unable to parse API response", Code: statusCode}
+	}
+
+	namesInterface, ok := result["names"].([]interface{})
```
nmanoogian: optional: Rather than fighting with `[]interface{}`, you might find it a little easier to marshal directly into the object shape that you're looking for:

```diff
diff --git a/pkg/http/api.go b/pkg/http/api.go
index 4b7d645..f879205 100644
--- a/pkg/http/api.go
+++ b/pkg/http/api.go
@@ -240,27 +240,15 @@ func GetSecretNames(host string, verifyTLS bool, apiKey string, project string,
 		return nil, Error{Err: err, Message: "Unable to fetch secret names", Code: statusCode}
 	}
 
-	var result map[string]interface{}
+	var result struct {
+		Names []string `json:"names"`
+	}
 	err = json.Unmarshal(response, &result)
 	if err != nil {
 		return nil, Error{Err: err, Message: "Unable to parse API response", Code: statusCode}
 	}
 
-	namesInterface, ok := result["names"].([]interface{})
-	if !ok {
-		return nil, Error{Err: fmt.Errorf("Unexpected type parsing secret names, expected []interface{}, got %T", result["names"]), Message: "Unable to parse API response", Code: statusCode}
-	}
-
-	var names []string
-	for _, value := range namesInterface {
-		v, ok := value.(string)
-		if !ok {
-			return nil, Error{Err: fmt.Errorf("Unexpected type parsing secret name, expected string, got %T", value), Message: "Unable to parse API response", Code: statusCode}
-		}
-		names = append(names, v)
-	}
-
-	return names, Error{}
+	return result.Names, Error{}
 }
 
 // UploadSecrets for specified project and config

```

pkg/http/api.go
```diff
@@ -171,17 +171,12 @@ func GetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
 	var params []queryParam
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "config", Value: config})
+	params = append(params, queryParam{Key: "include_dynamic_secrets", Value: strconv.FormatBool(includeDynamicSecrets)})
```
nmanoogian, line 174: Nice one!

pkg/cmd/secrets.go
```diff
@@ -538,8 +538,29 @@ func substituteSecrets(cmd *cobra.Command, args []string) {
...
+
+	// fetch secrets
 	dynamicSecretsTTL := utils.GetDurationFlag(cmd, "dynamic-ttl")
-	response, responseErr := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, nil, true, dynamicSecretsTTL)
+	response, responseErr := http.GetSecrets(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, secretNames, true, dynamicSecretsTTL)
```
nmanoogian: As it stands, if `secretNames` is `[]` (i.e. no names are found in the template), `GetSecrets` will request all secrets

pkg/cmd/secrets.go
nmanoogian: I'm worried about changing the default to this new behavior because it's a breaking change. I know we have a least a few users who write dynamic templates that don't reference secrets directly at all. For example,

```
{{ range $key, $value := . }}
{{ $key }}==={{ $value }}
{{ end }}
```

nmanoogian (changes requested):

pkg/cmd/secrets.go
Piccirello: Really glad you have context on usage in the real world. There's still an option to parse the template fields and, if all fields are of the format "{{.SECRET_NAME}}", then fetch those parsed secrets. I'm not sure this complexity is worth it though, and would be fine abandoning this specific feature. Wdyt?

nmanoogian: I think it'll be really tricky to get the heuristics correct for templates. For example, whitespace is allowed around the control chars (e.g. `{{ .SECRET_NAME }}`) and there are special whitespace consume indicators (e.g. `{{ .DOPPLER_CONFIG -}}`).

I think the safest option is to match the same behavior as `doppler run`. By default, all secrets are provided to the template and you can narrow that list manually with `--only-secrets`.

Piccirello: Agreed that parsing the template ourselves isn't viable. I'm going to hold off on supporting this in `doppler secrets substitute` for now.

pkg/cmd/secrets.go
Piccirello: Good call out - though thankfully no longer relevant with the dropping of template parsing functionality.

pkg/http/api.go
Piccirello: Much nicer! I went a bit boilerplate heavy, it seems.

Piccirello:
FYI the last commit was from #349, which I merged prematurely.

nmanoogian (approved):

PR #351, opened 2022-12-15
Only fetch requested secrets when using `run --only-secrets`

Piccirello:
This PR also adds support for fallback files when using name transformers.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -93,6 +93,20 @@ doppler run --mount secrets.json -- cat secrets.json`,
 
 		utils.RequireValue("token", localConfig.Token.Value)
 
+		if cmd.Flags().Changed("only-secrets") && len(secretsToInclude) == 0 {
+			utils.HandleError(fmt.Errorf("you must specify secrets when using --only-secrets"))
+		}
+
+		nameTransformerString := cmd.Flag("name-transformer").Value.String()
+		var nameTransformer *models.SecretsNameTransformer
+		if nameTransformerString != "" {
+			nameTransformer = models.SecretsNameTransformerMap[nameTransformerString]
+			if nameTransformer == nil || !nameTransformer.EnvCompat {
+				utils.HandleError(fmt.Errorf("invalid name transformer. Valid transformers are %s", validEnvCompatNameTransformersList))
+			}
+		}
+
+		const format = models.JSON
 		fallbackPath := ""
 		legacyFallbackPath := ""
 		metadataPath := ""
@@ -114,15 +128,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 				if cmd.Flags().Changed(flag) {
 					utils.LogWarning(fmt.Sprintf("--%s has no effect when the fallback file is disabled", flag))
 				}
-			}
-		}
-
-		nameTransformerString := cmd.Flag("name-transformer").Value.String()
-		var nameTransformer *models.SecretsNameTransformer
-		if nameTransformerString != "" {
-			nameTransformer = models.SecretsNameTransformerMap[nameTransformerString]
-			if nameTransformer == nil || !nameTransformer.EnvCompat {
-				utils.HandleError(fmt.Errorf("invalid name transformer. Valid transformers are %s", validEnvCompatNameTransformersList))
 			}
 		}
 
@@ -157,10 +162,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		}
 
 		if cmd.Flags().Changed("only-secrets") {
-			if len(secretsToInclude) == 0 {
-				utils.HandleError(fmt.Errorf("you must specify secrets when using --only-secrets"))
-			}
-
 			var err error
 			noExitOnMissingIncludedSecrets := cmd.Flags().Changed("no-exit-on-missing-only-secrets")
 			dopplerSecrets, err = controllers.SelectSecrets(dopplerSecrets, secretsToInclude)
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -468,19 +468,6 @@ func fetchSecrets(localConfig models.ScopedOptions, enableCache bool, fallbackOp
 			}
 		}
 
-		// TODO remove this when releasing CLI v4 (DPLR-435)
-		if fallbackOpts.legacyPath != "" && localConfig.EnclaveProject.Value != "" && localConfig.EnclaveConfig.Value != "" {
-			utils.LogDebug(fmt.Sprintf("Writing to legacy fallback file %s", fallbackOpts.legacyPath))
-			if err := utils.WriteFile(fallbackOpts.legacyPath, []byte(encryptedResponse), utils.RestrictedFilePerms()); err != nil {
-				utils.Log("Unable to write to legacy fallback file")
-				if fallbackOpts.exitOnWriteFailure {
-					utils.HandleError(err, "", strings.Join(writeFailureMessage(), "\n"))
-				} else {
-					utils.LogDebugError(err)
-				}
-			}
-		}
-
 		if enableCache {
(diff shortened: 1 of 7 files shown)


pkg/cmd/run.go
```diff
@@ -462,19 +462,6 @@ func fetchSecrets(localConfig models.ScopedOptions, enableCache bool, fallbackOp
...
-			utils.LogDebug(fmt.Sprintf("Writing to legacy fallback file %s", fallbackOpts.legacyPath))
-			if err := utils.WriteFile(fallbackOpts.legacyPath, []byte(encryptedResponse), utils.RestrictedFilePerms()); err != nil {
-				utils.Log("Unable to write to legacy fallback file")
-				if fallbackOpts.exitOnWriteFailure {
-					utils.HandleError(err, "", strings.Join(writeFailureMessage(), "\n"))
-				} else {
-					utils.LogDebugError(err)
-				}
-			}
-		}
```
Piccirello, line 481: This is technically a breaking change in that `doppler run` performed with the latest client won't generate a fallback file that CLI v3.14.0 (and below) can understand. Given that [v3.14.0](https://github.com/DopplerHQ/cli/releases/tag/3.14.0) shipped over 2 years ago, I think this risk is acceptable.

Kunamatata: Does this mean the CLI is moving to `v4`?

pkg/cmd/run.go
```diff
@@ -157,10 +162,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		}
 
 		if cmd.Flags().Changed("only-secrets") {
-			if len(secretsToInclude) == 0 {
-				utils.HandleError(fmt.Errorf("you must specify secrets when using --only-secrets"))
-			}
-
 			var err error
 			noExitOnMissingIncludedSecrets := cmd.Flags().Changed("no-exit-on-missing-only-secrets")
 			dopplerSecrets, err = controllers.SelectSecrets(dopplerSecrets, secretsToInclude)
```
Kunamatata: With the new functionality of downloading only specified secrets, we can re-organize this block and rework the `SelectSecrets` function. We no longer need to filter on the client side since the server only returns the specified secrets.

We might want to refactor `SelectSecrets`, remove the filter logic, rename the function to something like `detectedMissingSecrets` which takes the `secretsToInclude` and the result of the fetch call https://github.com/DopplerHQ/cli/blob/8ac1e7cfa3183c3885ed9bd2a99976883320016a/pkg/controllers/secrets.go#L243-L259

Maybe something along the lines of:
```go
func detectedMissingIncludedSecrets(secrets map[string]string, secretsToInclude []string) (map[string]string, error) {
	nonExistentSecretNames := []string{}

	for _, secretName := range secretsToInclude {
		if _, found := secrets[secretName]; !found {
			nonExistentSecretNames = append(nonExistentSecretNames, secretName)
		}
	}

	var err error
	if len(nonExistentSecretNames) > 0 {
		err = fmt.Errorf("the following secrets you are trying to include do not exist in your config:\n- %v", strings.Join(nonExistentSecretNames, "\n- "))
	}

	return secrets, err
}
```


```go
var dopplerSecrets = fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, 
nameTransformer, dynamicSecretsTTL, format, secretsToInclude)
// ...
if cmd.Flags().Changed("only-secrets") {
	res, err := detectMissingIncludedSecrets(dopplerSecrets, secretsToInclude)
        //...
}
```

Kunamatata (changes requested):
Really cool to see this @Piccirello! I think we can get rid of some of the logic that came with the previous work associated with `--only-secrets`

pkg/cmd/run.go
Piccirello: Good call, we can definitely remove this client-side filtering logic. Will refactor.

pkg/cmd/run.go
Piccirello: I think this is an acceptable level of risk so going to remain on v3. Major version upgrades shouldn't be a big deal, though they've traditionally been annoying as a lot of folks pin to a specific major version in their dockerfiles.

Though given that we've moved away from recommending the docker image as the default method (we recommend install.sh or using your OS's package manager), this probably isn't as big a deal as it used to be.

pkg/cmd/secrets.go
```diff
@@ -463,7 +463,7 @@ func downloadSecrets(cmd *cobra.Command, args []string) {
 			exitOnWriteFailure: exitOnWriteFailure,
 			passphrase:         fallbackPassphrase,
 		}
-		secrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL, format)
+		secrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL, format, nil)
```
nmanoogian, line 466: optional if irritating: I think this `nil` param goes with the next commit

pkg/controllers/fallback.go
```diff
@@ -42,13 +42,16 @@ func GenerateFallbackFileHash(token string, project string, config string, forma
 	if nameTransformer != nil {
 		parts = append(parts, nameTransformer.Type)
 	}
+	if len(secretNames) > 0 {
+		parts = append(parts, strings.Join(secretNames, ","))
```
nmanoogian: I think the answer to this question will be somewhat dependent on [server #5290](https://github.com/DopplerHQ/server/pull/5290) but would we want to dedupe and/or sort these?

nmanoogian (approved):

pkg/controllers/fallback.go
Piccirello: Great idea, that definitely seems worthwhile.

pkg/cmd/run.go
Piccirello: Updated this logic to only search for missing secrets

pkg/cmd/secrets.go
Piccirello: Fixed!

nmanoogian (dismissed):

Kunamatata (dismissed):
🎉

pkg/http/api.go
```diff
@@ -138,12 +138,15 @@ func RevokeAuthToken(host string, verifyTLS bool, token string) (map[string]inte
...
-func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string, config string, format models.SecretsFormat, nameTransformer *models.SecretsNameTransformer, etag string, dynamicSecretsTTL time.Duration) (int, http.Header, []byte, Error) {
+func DownloadSecrets(host string, verifyTLS bool, apiKey string, project string, config string, format models.SecretsFormat, nameTransformer *models.SecretsNameTransformer, etag string, dynamicSecretsTTL time.Duration, secrets []string) (int, http.Header, []byte, Error) {
 	var params []queryParam
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "config", Value: config})
 	params = append(params, queryParam{Key: "format", Value: format.String()})
 	params = append(params, queryParam{Key: "include_dynamic_secrets", Value: "true"})
+	if len(secrets) > 0 {
+		params = append(params, queryParam{Key: "secrets", Value: strings.Join(secrets, ",")})
+	}
```
Kunamatata, line 149: This might be a big refactor, but we could leverage the `url` stdlib go package for query params, URL encoding and more. It's the same idea as the `new URL(...)` from js (I made a similar comment on the notes PR: https://github.com/DopplerHQ/cli/pull/352#discussion_r1052618951

Piccirello: +1 that's a great call out. I do think that's a bigger refactor but will create a linear ticket for it. I'd love to also move to passing around a `context` object, which will help further consolidate a bunch of args we're shuffling around.

nmanoogian (approved):

PR #352, opened 2022-12-15
Support setting notes on secrets

Piccirello:
This PR adds a new command for setting notes on a secret: `doppler secrets notes set`. A separate command was added as adding this to `doppler secrets set` would mean that a user wouldn't be able to set a secret note without also setting the secret value.

This is in draft pending the server PR.

Closes ENG-1669.

### DIFF
--- a/pkg/cmd/configure.go
+++ b/pkg/cmd/configure.go
@@ -16,10 +16,8 @@ limitations under the License.
 package cmd
 
 import (
-	"bufio"
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
@@ -166,25 +164,15 @@ doppler configure set key=123 otherkey=456`,
 		options := map[string]string{}
 
 		if len(args) == 1 && !strings.Contains(args[0], "=") {
-			var input []string
-			scanner := bufio.NewScanner(os.Stdin)
-			// read input from stdin
-			for {
-				if ok := scanner.Scan(); !ok {
-					if e := scanner.Err(); e != nil {
-						utils.HandleError(e, "Unable to read input from stdin")
-					}
-
-					break
-				}
-
-				s := scanner.Text()
-				input = append(input, s)
+			value, err := utils.GetStdIn()
+			if err != nil {
+				utils.HandleError(err)
+			}
+			if value == nil {
+				utils.HandleError(errors.New("Unable to read input from stdin"))
 			}
 
-			key := args[0]
-			value := strings.Join(input, "\n")
-			options[key] = value
+			options[args[0]] = *value
 		} else if !strings.Contains(args[0], "=") {
 			options[args[0]] = args[1]
 		} else {
--- a/pkg/cmd/secrets_notes.go
+++ b/pkg/cmd/secrets_notes.go
@@ -16,10 +16,6 @@ limitations under the License.
 package cmd
 
 import (
-	"bufio"
-	"os"
-	"strings"
-
 	"github.com/DopplerHQ/cli/pkg/configuration"
 	"github.com/DopplerHQ/cli/pkg/http"
 	"github.com/DopplerHQ/cli/pkg/printer"
@@ -47,40 +43,23 @@ func setSecretNote(cmd *cobra.Command, args []string) {
 	utils.RequireValue("token", localConfig.Token.Value)
 
 	secret := args[0]
+	utils.RequireValue("secret", secret)
+
 	var note string
 	if len(args) > 1 {
 		note = args[1]
 	} else {
 		// read from stdin
-		hasData, e := utils.HasDataOnStdIn()
-		if e != nil {
-			utils.HandleError(e)
+		noteString, err := utils.GetStdIn()
+		if err != nil {
+			utils.HandleError(err)
 		}
-
-		// note must be supplied
-		if !hasData {
+		if noteString == nil {
 			utils.RequireValue("note", note)
 		}
 
-		var input []string
-		scanner := bufio.NewScanner(os.Stdin)
-		for {
-			if ok := scanner.Scan(); !ok {
-				if e := scanner.Err(); e != nil {
-					utils.HandleError(e, "Unable to read input from stdin")
-				}
-
-				break
-			}
-
-			s := scanner.Text()
-			input = append(input, s)
-		}
-
-		note = strings.Join(input, "\n")
+		note = *noteString
 	}
-
-	utils.RequireValue("secret", secret)
 
 	response, httpErr := http.SetSecretNote(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, localConfig.EnclaveProject.Value, localConfig.EnclaveConfig.Value, secret, note)
 	if !httpErr.IsNil() {
--- a/pkg/cmd/secrets_notes.go
+++ b/pkg/cmd/secrets_notes.go
@@ -1,0 +1,101 @@
+/*
+Copyright © 2020 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"bufio"
+	"os"
+	"strings"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/printer"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var secretsNotesCmd = &cobra.Command{
+	Use:   "notes",
+	Short: "Manage secret notes",
+	Args:  cobra.NoArgs,
+}
+
+var secretsNotesSetCmd = &cobra.Command{
+	Use:   "set [secret] [note]",
+	Short: "Set a note on a secret. The secret must exist. Notes can be passed via arg or via stdin.",
+	Args:  cobra.RangeArgs(1, 2),
+	Run:   setSecretNote,
+}
+
+func setSecretNote(cmd *cobra.Command, args []string) {
+	jsonFlag := utils.OutputJSON
+	localConfig := configuration.LocalConfig(cmd)
+
+	utils.RequireValue("token", localConfig.Token.Value)
+
+	secret := args[0]
+	var note string
+	if len(args) > 1 {
+		note = args[1]
+	} else {
+		// read from stdin
+		hasData, e := utils.HasDataOnStdIn()
+		if e != nil {
+			utils.HandleError(e)
+		}
+
+		// note must be supplied
+		if !hasData {
+			utils.RequireValue("note", note)
+		}
+
+		var input []string
+		scanner := bufio.NewScanner(os.Stdin)
+		for {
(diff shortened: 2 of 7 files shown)


pkg/http/api.go
```diff
@@ -231,6 +231,34 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
 	return computed, Error{}
 }
 
+// SetSecretNote for specified project and config
+func SetSecretNote(host string, verifyTLS bool, apiKey string, project string, config string, secret string, note string) (models.SecretNote, Error) {
```
Kunamatata, line 235: This might be a bigger refactor so feel free to let me know if it's out of scope.

What do you think of wrapping our HTTP client inside our own struct that would contain the host `localConfig := configuration.LocalConfig(cmd)` at instanciation time. 
This would make it so that we don't have to pass  `host` to `SetSecretNote` It shouldn't be the responsibility of `SetSecretNoet` to know about or worry about the host.

pkg/http/api.go
```diff
@@ -231,6 +231,34 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+	reqBody := map[string]interface{}{}
+	reqBody["secret"] = secret
+	reqBody["note"] = note
+	body, err := json.Marshal(reqBody)
```
Kunamatata: Do we want to type this struct fully instead of using `map[string]interface{}{}`?

pkg/http/api.go
```diff
@@ -231,6 +231,34 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+	if err != nil {
+		return models.SecretNote{}, Error{Err: err, Message: "Invalid secret note"}
+	}
+
+	var params []queryParam
+	params = append(params, queryParam{Key: "project", Value: project})
+	params = append(params, queryParam{Key: "config", Value: config})
+
+	statusCode, _, response, err := PostRequest(host, verifyTLS, apiKeyHeader(apiKey), "/v3/configs/config/secrets/note", params, body)
```
Kunamatata, line 245: Go has a similar mechanism to JavaScript for URLs that we could leverage:
```go
url, err := url.Parse("https://www.aurl.com")
if err != nil {
	//...
}
values := url.Query()
values.Add("key1", "value1")
values.Add("key2", "value2")
// values.Del
// values.Set
url.RawQuery = values.Encode()

log.Println(url.String())
```

Would this be a worthwhile refactor?

pkg/http/api.go
```diff
@@ -231,6 +231,34 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+	if err != nil {
+		return models.SecretNote{}, Error{Err: err, Message: "Unable to set secret note", Code: statusCode}
+	}
+
+	var result map[string]interface{}
+	err = json.Unmarshal(response, &result)
+	if err != nil {
+		return models.SecretNote{}, Error{Err: err, Message: "Unable to parse API response", Code: statusCode}
```
Kunamatata: Same comment, do we want to create a struct for responses instead of using `map[string]interface{}`?

pkg/http/api.go
```diff
@@ -231,6 +231,34 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+	}
+
+	return models.SecretNote{Secret: result["secret"].(string), Note: result["note"].(string)}, Error{}
```
Kunamatata: Just want to confirm that if a secret does not have a note we send back:
```javascript
{
  secret: "SECRET",
  note: ""
}
```

pkg/cmd/secrets_notes.go
```diff
@@ -0,0 +1,66 @@
...
+		printer.SecretNote(response, jsonFlag)
+	}
+}
+
+func init() {
+	secretsNotesSetCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
+	secretsNotesSetCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
+	secretsNotesCmd.AddCommand(secretsNotesSetCmd)
+
+	secretsCmd.AddCommand(secretsNotesCmd)
```
Kunamatata, line 79: 2 questions:

- Do we want the notes command to live under the secrets command. Notes are tied to secrets but the command feels kind of off:
`doppler secrets notes set MY_SECRET "Hello, World"`
- Should I be able to make a multiline note from the CLI? Adding `\n` in my note did not translate into a new line.

Kunamatata:
Probably two possible other tickets:
- Do we want a delete endpoint for notes, or would you expect users to call `... notes set MY_SECRET ""` ?
- Do we want to allow users to see notes for a given secret via the CLI?

pkg/http/api.go
nmanoogian: +1, it looks like you already have `SecretNote` which has JSON struct tags

nmanoogian (dismissed):
Looking good modulo @Kunamatata's struct comments!

nmanoogian:
> * Do we want a delete endpoint for notes, or would you expect users to call `... notes set MY_SECRET ""` ?

Great point. It looks like `doppler secrets notes set MY_SECRET ""` doesn't actually work:

```
$ doppler secrets notes set PRIVATE_KEY ""
Error: accepts 2 arg(s), received 1
Usage:
  doppler secrets notes set [secret] [note] [flags]
```

Kunamatata:
@nmanoogian It does work for me:
image

nmanoogian:
> @nmanoogian It does work for me: image

Interesting! Can you try it with the built binary?

pkg/http/api.go
Piccirello: This is definitely a larger refactor but I agree there's a better way to be passing this data. At a minimum we should move to using a context, which could contain this and other configuration.

pkg/http/api.go
Piccirello: Discussed elsewhere: https://github.com/DopplerHQ/cli/pull/351#discussion_r1052816339

pkg/cmd/secrets_notes.go
Piccirello: - I couldn't think of a better location for it, but open to suggestions. I don't think it should live at the root `doppler notes`
- Indeed you should. I'll add support for reading the secret note from stdin, which will allow for multiline values.

Piccirello:
> Interesting! Can you try it with the built binary?

I added support for this at some point in a force push, so you may be using an older binary.

Piccirello:
> Probably two possible other tickets:
> 
>     * Do we want a delete endpoint for notes, or would you expect users to call `... notes set MY_SECRET ""` ?
> 
>     * Do we want to allow users to see notes for a given secret via the CLI?

I don't think we need an explicit delete endpoint - users can set a blank note like you mentioned.

You can see notes for a specific secret using `doppler secrets get`, or for all secrets using `doppler secrets`. I don't think we need an explicit `doppler secrets notes get` command just yet, though we may add one in the future.

pkg/cmd/secrets_notes.go
Piccirello: Added multiline note support https://github.com/DopplerHQ/cli/compare/11a1bed0026e6404652a66a07fbd33c410367cf4..d0ae528280d509092c15d5e2e47b9d69644e0dec

To test: `echo -e "foo\nbar" | ./doppler secrets notes set S1`

Kunamatata: I don't have a better suggestion of where the command should live. Like you wrote, the root isn't really where it should be either. Let's go with this for now since there's nowhere else to really put it.

Nice job on the multiline :D

Kunamatata:
@nmanoogian It also works with the binary when running `go build` it names it as `cli` so `./cli secrets note set MY_SECRET ""` sets an empty note.

pkg/http/api.go
Kunamatata: If you tackle this in the future or make a linear ticket I'd love to be involved and see how we approach this. I wonder if context is the right spot to put these things.

pkg/cmd/secrets_notes.go
```diff
@@ -0,0 +1,101 @@
...
+				}
+
+				break
+			}
+
+			s := scanner.Text()
+			input = append(input, s)
+		}
+
+		note = strings.Join(input, "\n")
```
Kunamatata: Do we have any other spots in the CLI that would benefit from taking stdin as input? If so, do you think it would be valuable to extract this out of `setSecretNote` to make it reusable by other commands?

Piccirello: We currently support taking input via stdin in two other commands: `doppler secrets set` and `doppler configure set`. `doppler secrets set` would be difficult to refactor as it has a lot of custom logic, though `doppler configure set` is very similar to the new logic. I'll see if I can extract this logic for `doppler secrets notes set` and `doppler configure set`.

Kunamatata: Sweet! We can also do a follow-up refactor ticket for these if that sounds more reasonable!

Piccirello: Just pushed up a commit for this

Kunamatata (dismissed):
Awesome!

pkg/http/api.go
Piccirello: Confirmed, this endpoint always sends back a note of type `string`. When there's no note, it'll be an empty string.

Kunamatata:
@Piccirello quick follow up on this, https://github.com/DopplerHQ/cli/pull/352#discussion_r1052621458 you put a thumbs up, where do you stand?

EDIT: posted the comment right when you pushed! Nice

Piccirello:
> @Piccirello quick follow up on this, [#352 (comment)](https://github.com/DopplerHQ/cli/pull/352#discussion_r1052621458) you put a thumbs up, where do you stand?
> 
> EDIT: posted the comment right when you pushed! Nice

All fixed now. Should be good for a re-review.

pkg/http/api.go
```diff
@@ -231,6 +231,31 @@ func SetSecrets(host string, verifyTLS bool, apiKey string, project string, conf
...
+		return models.SecretNote{}, Error{Err: err, Message: "Unable to set secret note", Code: statusCode}
+	}
+
+	var secretNote models.SecretNote
+	err = json.Unmarshal(response, &secretNote)
+	if err != nil {
+		return models.SecretNote{}, Error{Err: err, Message: "Unable to parse API response", Code: statusCode}
+	}
+
+	return secretNote, Error{}
```
Kunamatata, line 256: @Piccirello What is the reason for returning an `Error{}` instead of `nil`?

Piccirello: The return type is defined as `Error`, which is incompatible with `nil`. You must return an empty struct and check it with `if !httpErr.IsNil()`. We use this pattern throughout.

If that's your only remaining comment, please approve :)

Kunamatata (dismissed):

pkg/cmd/configure.go
```diff
@@ -166,25 +164,15 @@ doppler configure set key=123 otherkey=456`,
...
-					if e := scanner.Err(); e != nil {
-						utils.HandleError(e, "Unable to read input from stdin")
-					}
-
-					break
-				}
-
-				s := scanner.Text()
-				input = append(input, s)
+			value, ok, err := utils.GetStdIn()
```
nmanoogian: When the var is named `ok`, this util's signature feels funky; the presence of `err` seems like it should indicate whether or not the read was "ok". However, if this were named "hasData", I think it would make sense. Thoughts on renaming the var here (and at the other callsite)? Alternatives include:

- Changing the signature to `(string, error)` and returning `"", errors.New("Unable to read input from stdin")` directly from `GetStdIn` if there is no data
- Changing the signature to `(*string, error)` and returning `nil, nil` if there is no data

nmanoogian (dismissed):

pkg/cmd/configure.go
Piccirello: I liked your last suggestion of changing the return signature to `(*string, error)` as it seemed the least ambiguous. Change has been pushed!

For posterity, I didn't go with the second suggestion as I want the caller to be able to determine whether a lack of data on stdin is an error.

nmanoogian (approved):

Kunamatata (approved):

PR #353, opened 2022-12-15
Add anonymous analytics for tracking command usage

Piccirello:
This allows us to see what commands people are most frequently using. Analytics are enabled by default and are completely anonymous.

Data we collect:
- command name
- os
- architecture
- CLI version

To opt out: run `doppler analytics disable`.

Closes ENG-735.

### DIFF
--- a/pkg/cmd/analytics.go
+++ b/pkg/cmd/analytics.go
@@ -1,0 +1,88 @@
+/*
+Copyright © 2022 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/printer"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var analyticsCmd = &cobra.Command{
+	Use:   "analytics",
+	Short: "Manage anonymous analytics",
+	Args:  cobra.NoArgs,
+}
+
+var analyticsStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Check whether anonymous analytics are enabled",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		if utils.OutputJSON {
+			printer.JSON(map[string]bool{"enabled": configuration.IsAnalyticsEnabled()})
+		} else {
+			adjective := "enabled"
+			if !configuration.IsAnalyticsEnabled() {
+				adjective = "disabled"
+			}
+			utils.Log(fmt.Sprintf("Anonymous analytics are currently %s", adjective))
+		}
+	},
+}
+
+var analyticsEnableCmd = &cobra.Command{
+	Use:   "enable",
+	Short: "Enable anonymous analytics",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		configuration.EnableAnalytics()
+
+		if utils.OutputJSON {
+			printer.JSON(map[string]bool{"enabled": true})
+		} else {
+			utils.Log("Anonymous analytics have been enabled")
+		}
+	},
+}
+
+var analyticsDisableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable anonymous analytics",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		configuration.DisableAnalytics()
+
+		if utils.OutputJSON {
+			printer.JSON(map[string]bool{"enabled": false})
+		} else {
+			utils.Log("Anonymous analytics have been disabled")
+		}
+	},
+}
+
+func init() {
+	analyticsCmd.AddCommand(analyticsStatusCmd)
+
+	analyticsCmd.AddCommand(analyticsEnableCmd)
+
+	analyticsCmd.AddCommand(analyticsDisableCmd)
+
+	rootCmd.AddCommand(analyticsCmd)
+}
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -18,6 +18,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
@@ -31,6 +32,7 @@ import (
 	"gopkg.in/gookit/color.v1"
 )
 
+var wg *sync.WaitGroup
 var printConfig = false
 
 var rootCmd = &cobra.Command{
@@ -41,6 +43,9 @@ var rootCmd = &cobra.Command{
 		loadFlags(cmd)
 		configuration.Setup()
 		configuration.LoadConfig()
+
+		wg.Add(1)
+		go controllers.CaptureCommand(wg, cmd.CommandPath())
 
 		if utils.Debug && utils.Silent {
 			utils.LogWarning("--silent has no effect when used with --debug")
@@ -184,7 +189,15 @@ func Execute() {
 		}
 	}()
 
-	if err := rootCmd.Execute(); err != nil {
+	// initialize the wait group before executing the command
+	wg = new(sync.WaitGroup)
+
+	err := rootCmd.Execute()
+
+	// wait for group before checking error
+	wg.Wait()
+
+	if err != nil {
 		os.Exit(1)
 	}
 }
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -67,7 +67,7 @@ var rootCmd = &cobra.Command{
 		// --plain doesn't normally affect logging output, but due to legacy reasons it does here
 		// also don't want to display updates if user doesn't want to be prompted (--no-prompt/--no-interactive)
 		if isTTY && utils.CanLogInfo() && !plain && canPrompt {
-			checkVersion(cmd.CommandPath())
+			checkVersion(wg, cmd.CommandPath())
 		}
 	},
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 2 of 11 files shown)


pkg/cmd/root.go
```diff
@@ -42,6 +43,10 @@ var rootCmd = &cobra.Command{
 		configuration.Setup()
 		configuration.LoadConfig()
 
+		if configuration.IsAnalyticsEnabled() {
+			controllers.CaptureCommand(strings.ReplaceAll(cmd.CommandPath(), " ", "."))
```
Kunamatata: Should it be the responsibility of `controllers.CaptureCommand(...)` to format the `cmd.CommandPath()` via `strings.ReplaceAll(cmd.CommandPath(), " ", ".")` instead of the caller doing it?

pkg/http/analytics.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/DopplerHQ/cli/pkg/utils"
+)
+
+func CaptureCommand(command string) ([]byte, Error) {
+	postBody := map[string]interface{}{"command": command}
```
Kunamatata, line 26: Do we want to have a struct for these instead of using `map[string]interface{}`? 
`map[string]interface{}` is comparable to typing with `any` in typescript, we lose the ability to rely on types for our objects.

```go
type CaptureCommand struct {
	Command string `json:"command"`
}
```

pkg/cmd/analytics.go
```diff
@@ -0,0 +1,88 @@
...
+	Args:  cobra.NoArgs,
+}
+
+var analyticsStatusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Check whether anonymous analytics are enabled",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		if utils.OutputJSON {
+			printer.JSON(map[string]bool{"enabled": configuration.IsAnalyticsEnabled()})
```
Kunamatata, line 39: Same comment about leveraging a struct instead of map[string]bool:
```go
type AnalyticsStatus struct {
	Enabled bool `json:"enabled"`
}
```

pkg/controllers/analytics.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"github.com/DopplerHQ/cli/pkg/http"
+)
+
+// This package collects anonymous analytics for the purpose of improving the Doppler CLI
+
+func CaptureCommand(command string) {
+	// wait for analytics command to complete before exiting
+	ch := make(chan bool, 1)
```
nmanoogian: What's happening with this channel here? Won't this function just block until `CaptureCommand` finishes?

pkg/http/analytics.go
```diff
@@ -0,0 +1,39 @@
...
+	body, err := json.Marshal(postBody)
+	if err != nil {
+		return nil, Error{Err: err, Message: "Unable to marshall command"}
```
nmanoogian: nit: `marshal` (just one L)

pkg/http/analytics.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	utils.LogDebug(fmt.Sprintf("Sending anonymous analytics payload: '%s'", body))
+
+	_, _, resp, err := PostRequest("https://cli.doppler.com", true, map[string]string{"Content-Type": "application/json"}, "/v1/analytics", nil, body)
+	if err != nil {
+		return nil, Error{Err: err, Message: "Unable to fetch changelog"}
```
nmanoogian: nit: copy-pasta

pkg/controllers/analytics.go
Kunamatata: https://pkg.go.dev/sync#WaitGroup Might be a good feature to leverage. 
`wg.Wait()` could happen as the last line of the `Execute()` function in `root.go` and allow wg to be accessed by goroutines in our code.
The first line of the `CaptureCommand` function could be `wg.Add()` and the first line inside the goroutine function `defer wg.Done()`

We can then get rid of the channel. This will ensure that the `Execute()` function does not end (until the wait group counter is done to 0) which would end the main thread and end all goroutines.

EDIT:
I just thought about this but we can potentially also have a wait group at the callsite of `CaptureCommand` (`root.go`) and pass the wait group inside the CaptureCommand.

Piccirello: This indeed was not working as I had intended. I've switched to using a WaitGroup for this.

pkg/http/analytics.go
Piccirello: Since we immediately marshal to json on the next line, I'm not sure that the struct gets us anything.

pkg/cmd/analytics.go
Piccirello: Here too - we can add a struct, but I don't think we'd really get any benefit.

pkg/controllers/analytics.go
```diff
@@ -30,3 +30,9 @@ func CaptureCommand(wg *sync.WaitGroup, command string) {
 	command = strings.ReplaceAll(command, " ", ".")
 	http.CaptureCommand(command)
 }
+
+func CaptureEvent(wg *sync.WaitGroup, event string) {
+	defer wg.Done()
+
+	http.CaptureEvent(event)
```
nmanoogian: nit: To be very explicit that we're ignoring these returned values (and swallowing errors), it might be nice to assign the returned values to underscore:

```diff
diff --git a/pkg/controllers/analytics.go b/pkg/controllers/analytics.go
index 6d07189..4637176 100644
--- a/pkg/controllers/analytics.go
+++ b/pkg/controllers/analytics.go
@@ -28,11 +28,11 @@ func CaptureCommand(wg *sync.WaitGroup, command string) {
 	defer wg.Done()
 
 	command = strings.ReplaceAll(command, " ", ".")
-	http.CaptureCommand(command)
+	_, _ = http.CaptureCommand(command)
 }
 
 func CaptureEvent(wg *sync.WaitGroup, event string) {
 	defer wg.Done()
 
-	http.CaptureEvent(event)
+	_, _ = http.CaptureEvent(event)
 }


```

pkg/http/analytics.go
```diff
@@ -37,3 +37,19 @@ func CaptureCommand(command string) ([]byte, Error) {
 	}
 	return resp, Error{}
 }
+
+func CaptureEvent(event string) ([]byte, Error) {
+	postBody := map[string]interface{}{"event": event}
+	body, err := json.Marshal(postBody)
+	if err != nil {
+		return nil, Error{Err: err, Message: "Unable to marshal command"}
```
nmanoogian: nit: "event"

pkg/cmd/root.go
```diff
@@ -101,6 +101,9 @@ func checkVersion(command string) {
 		return
 	}
 
+	wg.Add(1)
+	go controllers.CaptureEvent(wg, "VersionCheck")
```
nmanoogian, line 103: Should these be wrapped with `EnableAnalytics` checks?

nmanoogian (dismissed):

pkg/http/analytics.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/DopplerHQ/cli/pkg/utils"
+)
+
+func CaptureCommand(command string) ([]byte, Error) {
+	postBody := map[string]interface{}{"command": command}
+	body, err := json.Marshal(postBody)
```
dp-franklin, line 27: Is there anything else we might want to capture that isn't intrusive?
Ideas:
- True/false: is `api-host` set to `https://api.doppler.com`?
- True/false: is `token` set?
- True/false: is the `scope` set to `/`?

tests/e2e/analytics.sh
```diff
@@ -0,0 +1,59 @@
...
+beforeEach
+
+# analytics defaults to enabled
+status="$("$DOPPLER_BINARY" analytics status --configuration=./temp-config --json)"
+[[ "$status" == '{"enabled":true}' ]] || error "ERROR: analytics not enabled"
+
+beforeEach
+
+# analytics defaults to enabled after clearing config
+"$DOPPLER_BINARY" configure reset --configuration=./temp-config --yes
```
dp-franklin, line 48: Nit: I think it would be better to test the `doppler configure reset` command *without* first deleting the existing temp-config file in `beforeEach`

pkg/cmd/root.go
```diff
@@ -31,6 +32,7 @@ import (
 	"gopkg.in/gookit/color.v1"
 )
 
+var wg *sync.WaitGroup
```
dp-franklin, line 35: Shouldn't `wg.Wait()` be called at some point?

dp-franklin (dismissed):

pkg/cmd/root.go
Piccirello: Good catch. I'm going to move the `IsAnalyticsEnabled` check into the analytics controller to avoid this issue in the future.

pkg/http/analytics.go
Piccirello: The scope one is particularly interesting. I have wondered how frequently users are specifying a scope rather than using the default global scope (specifically when logging in and when running setup). I'll create a separate ticket for this.

pkg/cmd/root.go
Piccirello: We call `Wait()` at the bottom of this file in `Execute()`

tests/e2e/analytics.sh
Piccirello: Good call, I'll add an additional test to verify `reset` works as expected when there's an existing config with analytics disabled.

pkg/cmd/root.go
dp-franklin: Ah, I see now. Nevermind.

pkg/controllers/analytics.go
Piccirello: Ended up debug logging the error message

pkg/http/analytics.go
Piccirello: Created ENG-5666

dp-franklin (approved):

pkg/http/analytics.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return nil, Error{Err: err, Message: "Unable to marshal command"}
+	}
+
+	utils.LogDebug(fmt.Sprintf("Sending anonymous analytics payload: '%s'", body))
+
+	_, _, resp, err := PostRequest("https://cli.doppler.com", true, map[string]string{"Content-Type": "application/json"}, "/v1/analytics", nil, body)
```
dp-franklin, line 34: Question: is 10 seconds a reasonable amount timeout for sending analytics? I think it is but I'd like to get a second opinion. The concern is that this could prevent the CLI from exiting in a timely manner for an otherwise simple command.

dp-franklin (approved):

nmanoogian (approved):

pkg/http/analytics.go
Piccirello: 10 seconds is likely longer than we need, especially because we don't actually care about the response. This will require a bit of rework given the current structure of our http package, so I'd opt for improving this in a follow up PR.

PR #355, opened 2023-01-09
Update install instructions for Debian and Ubuntu

HarunOr:
I have updated the installation instructions for Debian and Ubuntu, since apt-key has been deprecated and will last be available in Debian 11 and Ubuntu 22.04 (as per the [Debian manpage](https://manpages.debian.org/testing/apt/apt-key.8.en.html)).


Piccirello:
I've put up #399 to address this. Thank you for bringing this to our attention!

PR #360, opened 2023-01-16
chore: log request before performing request

Piccirello:
This makes it easier to see what's happening. In the event the server takes a while to respond, it'll be more clear that a request has been sent, rather than the CLI appearing like it's hanging.

Closes ENG-5675.

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -207,6 +207,8 @@ func performRequest(req *http.Request, verifyTLS bool, params []queryParam) (int
 		Proxy:             http.ProxyURL(proxyUrl),
 	}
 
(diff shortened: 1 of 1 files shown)


nmanoogian (approved):
Just calling out that we're moving this out of the `retry` block, so we'll no longer see this line for every attempt. I think that's fine 👍

Piccirello:
> Just calling out that we're moving this out of the `retry` block, so we'll no longer see this line for every attempt. I think that's fine 👍

It seems for any retry we do at least log an error, so we won't be completely in the dark.

Naml21:
fatihnamlıdurum.net

PR #354, opened 2022-12-20
chore: use url library for constructing URLs

Piccirello:
Closes ENG-5590.

PR #363, opened 2023-01-30
Capture timing metrics when performing update

Piccirello:
This change allows us to see how many installs succeed/fail, as well as how long the update process takes.

### DIFF
--- a/pkg/global/waitgroup.go
+++ b/pkg/global/waitgroup.go
@@ -1,0 +1,23 @@
+/*
+Copyright © 2020 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package global
+
+import "sync"
+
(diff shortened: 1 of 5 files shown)


pkg/global/waitgroup.go
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package global
+
+import "sync"
+
+var WaitGroup *sync.WaitGroup
```
nmanoogian, line 23: Given that it's global now, it might be handy to add a comment for `WaitGroup` explaining how producers and consumers are intended to use it

pkg/http/analytics.go
```diff
@@ -36,7 +36,7 @@ func CaptureCommand(command string) ([]byte, Error) {
 		return nil, Error{Err: err, Message: "Unable to generate url"}
 	}
 
-	_, _, resp, err := PutRequest(url, true, map[string]string{"Content-Type": "application/json"}, body)
```
nmanoogian, line 39: 😢 I can't believe [I missed this](https://github.com/DopplerHQ/cli/pull/354#discussion_r1085429932)

nmanoogian (dismissed):

pkg/global/waitgroup.go
Piccirello: I've added a comment noting the basic. Lmk what you think

Piccirello:
@nmanoogian I've pushed up an additional commit to simplify the analytics interface. This makes it so that the caller doesn't need to worry about goroutines or adding to the wait group.

nmanoogian (approved):
Nice change 👍 The new interface looks much easier to use

PR #364, opened 2023-02-01
Support pagination when listing projects, environments, and configs

Piccirello:


nmanoogian (approved):
Looking good!

One clarification: `ValidArgsFunction: projectIDsValidArgs` (and similar) uses the number=100, page=1 defaults. I assume that this is just used for autocompletion and not actual parameter validation? When testing locally, I changed the default number to 5 and I was still able to do `doppler projects get foo` where foo was outside the first page. Just wanted to check though.

Piccirello:
> Looking good!
> 
> One clarification: `ValidArgsFunction: projectIDsValidArgs` (and similar) uses the number=100, page=1 defaults. I assume that this is just used for autocompletion and not actual parameter validation? When testing locally, I changed the default number to 5 and I was still able to do `doppler projects get foo` where foo was outside the first page. Just wanted to check though.

Indeed, those functions are only used for autocomplete. If the user's arg doesn't appear in that list, the command will still execute. From the cobra docs:

> ValidArgsFunction is an optional function that provides valid non-flag arguments for shell completion. It is a dynamic version of using ValidArgs. Only one of ValidArgs and ValidArgsFunction can be used for a command.

PR #366, opened 2023-02-14
Fix 'doppler configs delete' and 'doppler environments delete' commands

nmanoogian:
...by passing query params to API function.

This fixes a regression introduced in bd094d1e7023174c24e58fbf1647ebe87bebce95

Closes ENG-5849

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -38,9 +38,17 @@ builds:
 
 archives:
 -
-  replacements:
-    darwin: macOS
-    386: i386
+  rlcp: true
+  name_template: >-
+    {{ .ProjectName }}_
+    {{- .Version }}_
+    {{- if eq .Os "darwin" }}macOS
+    {{- else }}{{ .Os }}{{ end }}_
+    {{- if eq .Arch "386" }}i386
+    {{- else }}{{ .Arch }}{{ end }}
+    {{- with .Arm }}v{{ . }}{{ end }}
+    {{- with .Mips }}_{{ . }}{{ end }}
+    {{- if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}
   format_overrides:
     - goos: windows
       format: zip
@@ -66,6 +74,7 @@ checksum:
   algorithm: sha256
 
 source:
+  rlcp: true
   enabled: true
   name_template: '{{ .ProjectName }}_{{ .Version }}_src'
   format: 'tar.gz'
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -38,9 +38,14 @@ builds:
...
-    darwin: macOS
-    386: i386
+  rlcp: true
+  name_template: >-
+    {{ .ProjectName }}_
+    {{- if eq .Os "darwin" }}macOS
+    {{- else }}{{ .Os }}{{ end }}_
+    {{- if eq .Arch "386" }}i386
+    {{- else }}{{ .Arch }}{{ end }}
+    {{- if .Arm }}v{{ .Arm }}{{ end }}
```
Piccirello: Will this `Arm` line change the existing naming scheme?

Piccirello (commented):
🤦‍♂️

.goreleaser.yml
nmanoogian: Good callout, I meant to add a comment that I need to test this. The goreleaser docs said that these templates would be equivalent but I just did some testing locally with `goreleaser release --snapshot` and I'm getting different artifacts before and after this change. I'll need to take a closer look at this

nmanoogian: I force pushed a change and I think I've got it!

I pulled the defaults from https://goreleaser.com/customization/nfpm/?h=nfpms and https://goreleaser.com/customization/archive/?h=archives to make the necessary modifications.

I ran `rm -rf dist && goreleaser release --snapshot; (find dist | sort)` to get the generated list of artifacts and they look identical apart from the version hashes 👍 

Screenshot 2023-02-16 at 5 19 50 PM

Piccirello: Nice verification! I much preferred goreleaser's old method for defining this, but so is life...

pkg/http/api.go
```diff
@@ -685,7 +685,7 @@ func DeleteEnvironment(host string, verifyTLS bool, apiKey string, project strin
 	params = append(params, queryParam{Key: "project", Value: project})
 	params = append(params, queryParam{Key: "environment", Value: environment})
 
-	url, err := generateURL(host, "/v3/environments/environment", nil)
+	url, err := generateURL(host, "/v3/environments/environment", params)
```
Piccirello, line 688: I manually re-reviewed my PR that broke these and couldn't find any other instances

Piccirello (approved):

pkg/http/api.go
nmanoogian: Cool, same here -- I found `configs` by testing and during re-review I found `environments`

.goreleaser.yml
nmanoogian: Agreed 😞

PR #367, opened 2023-02-16
Allow setting config dir with DOPPLER_CONFIG_DIR

watsonian:
In some situations (e.g., a devcontainer that's being built), it can be advantageous to configure a specific config directory via environment variables rather than having to specify it with every CLI command. This allows you to set `DOPPLER_CONFIG_DIR` to set that value. Passing in `--config-dir` will still override this if needed.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -150,6 +150,15 @@ func loadFlags(cmd *cobra.Command) {
 	configuration.Scope = normalizedScope
 
 	configuration.CanReadEnv = !utils.GetBoolFlag(cmd, "no-read-env")
+
+	// User Config Dir
+	if configuration.CanReadEnv {
+		userConfigDir := os.Getenv("DOPPLER_CONFIG_DIR")
+		if userConfigDir != "" {
+			utils.Log(valueFromEnvironmentNotice("DOPPLER_CONFIG_DIR"))
+			configuration.SetConfigDir(userConfigDir)
+		}
+	}
 	configuration.SetConfigDir(utils.GetPathFlagIfChanged(cmd, "config-dir", configuration.UserConfigDir))
 	configuration.UserConfigFile = utils.GetPathFlagIfChanged(cmd, "configuration", configuration.UserConfigFile)
 	http.UseTimeout = !utils.GetBoolFlag(cmd, "no-timeout")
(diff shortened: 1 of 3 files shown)


pkg/cmd/root.go
```diff
@@ -150,6 +150,15 @@ func loadFlags(cmd *cobra.Command) {
 	configuration.Scope = normalizedScope
 
 	configuration.CanReadEnv = !utils.GetBoolFlag(cmd, "no-read-env")
+
+	// User Config Dir
+	if configuration.CanReadEnv {
+		userConfigDir := os.Getenv("DOPPLER_CONFIG_DIR")
+		if userConfigDir != "" {
+			utils.LogDebug(fmt.Sprintf("DOPPLER_CONFIG_DIR setting config dir to %s", userConfigDir))
```
Piccirello: You can also use `valueFromEnvironmentNotice` here

pkg/configuration/config.go
```diff
@@ -234,6 +234,18 @@ func LocalConfig(cmd *cobra.Command) models.ScopedOptions {
...
+	if flagSet || localConfig.UserConfigDir.Value == "" {
+		localConfig.UserConfigDir.Value = cmd.Flag("config-dir").Value.String()
+		localConfig.UserConfigDir.Scope = "/"
+
+		if flagSet {
+			localConfig.UserConfigDir.Source = models.FlagSource.String()
+		} else {
+			localConfig.UserConfigDir.Source = models.DefaultValueSource.String()
+		}
+	}
```
Piccirello: Does it make sense to include the config dir in the configuration output? We don't do this for other values like `config-file`, as we only print values that can be stored in the actual config file. If anything, I'd probably just add the config dir to the output of the `doppler configure debug` command, since we also print the path to the config file.

watsonian: I had just gone through trying to reproduce the stuff that existed for other settings that were pulled from env-vars. As long as this keeps showing up in `doppler configure debug` (which it does) after removing this, seems fine to me! I agree that it shouldn't be stored in a config file somewhere (wasn't actually clear that this is what was being done here).

Piccirello: You can take the same approach as `DOPPLER_PASSPHRASE`. Thus you shouldn't need any changes to `configuration/config.go` or `models/config.go`.

watsonian: I tried backing out the changes in `models/config.go`, but then it stopped showing up in `doppler configure debug` output. That does seem like the sort of thing we'd want to show up there. Maybe there's some subset of what's being done in `models/config.go` that it can be paired down to?

Piccirello: The table output by `doppler configure debug` is exclusively for values that can be stored in the config file. Above the table, we also print the path of the config file that's used. You'll want to add the config dir there, which won't require modifying `models/config.go`.

watsonian: Ah, I see. I'm following now. Pushed these changes up!

Piccirello (approved):

PR #368, opened 2023-02-16
Allow the passphrase to be set via DOPPLER_PASSPHRASE

watsonian:
This adds the ability to set the passphrase to be used when decrypting a fallback file via the `DOPPLER_PASSPHRASE` environment variable rather than relying solely on the `--passphrase` flag. This can help reduce the chance that the passphrase will be leaked in bash histories or by viewing process listings.

Closes ENG-5148.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -582,6 +582,14 @@ func getPassphrase(cmd *cobra.Command, flag string, config models.ScopedOptions)
 		return cmd.Flag(flag).Value.String()
 	}
 
+	if configuration.CanReadEnv {
+		passphrase := os.Getenv("DOPPLER_PASSPHRASE")
+		if passphrase != "" {
+			utils.Print(valueFromEnvironmentNotice("DOPPLER_PASSPHRASE"))
+			return passphrase
+		}
+	}
+
 	if config.EnclaveProject.Value != "" && config.EnclaveConfig.Value != "" {
 		return fmt.Sprintf("%s:%s:%s", config.Token.Value, config.EnclaveProject.Value, config.EnclaveConfig.Value)
 	}
--- a/tests/e2e/run-fallback.sh
+++ b/tests/e2e/run-fallback.sh
@@ -66,6 +66,17 @@ beforeEach
 
 beforeEach
 
+# test 'run' respects custom passphrase from environment
+DOPPLER_PASSPHRASE=123456 "$DOPPLER_BINARY" run -- echo -n > /dev/null
+# ensure default passphrase fails
+"$DOPPLER_BINARY" run --fallback-only -- echo -n > /dev/null 2>&1 && (echo "ERROR: --passphrase flag is not respected (1)" && exit 1)
+# test decryption with custom passphrase flag
+"$DOPPLER_BINARY" run --fallback-only --passphrase=123456 -- echo -n > /dev/null || (echo "ERROR: --passphrase flag is not respected (2)" && exit 1)
+# test decryption with custom passphrase from environment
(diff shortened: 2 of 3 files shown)


tests/e2e/run-fallback.sh
```diff
@@ -66,6 +66,17 @@ beforeEach
...
 beforeEach
 
+# test 'run' respects custom passphrase from environment
+DOPPLER_PASSPHRASE=123456 "$DOPPLER_BINARY" run -- echo -n > /dev/null
+# ensure default passphrase fails
+"$DOPPLER_BINARY" run --fallback-only -- echo -n > /dev/null 2>&1 && (echo "ERROR: --passphrase flag is not respected" && exit 1)
+# test decryption with custom passphrase flag
+"$DOPPLER_BINARY" run --fallback-only --passphrase=123456 -- echo -n > /dev/null || (echo "ERROR: --passphrase2 flag is not respected" && exit 1)
+# test decryption with custom passphrase from environment
+DOPPLER_PASSPHRASE=123456 "$DOPPLER_BINARY" run --fallback-only -- echo -n > /dev/null || (echo "ERROR: --passphrase3 flag is not respected" && exit 1)
```
Piccirello:
```suggestion
"$DOPPLER_BINARY" run --fallback-only --passphrase=123456 -- echo -n > /dev/null || (echo "ERROR: --passphrase flag is not respected" && exit 1)
# test decryption with custom passphrase from environment
DOPPLER_PASSPHRASE=123456 "$DOPPLER_BINARY" run --fallback-only -- echo -n > /dev/null || (echo "ERROR: --passphrase flag is not respected" && exit 1)
```

pkg/cmd/run.go
```diff
@@ -582,6 +582,14 @@ func getPassphrase(cmd *cobra.Command, flag string, config models.ScopedOptions)
 		return cmd.Flag(flag).Value.String()
 	}
 
+	if configuration.CanReadEnv {
+		passphrase := os.Getenv("DOPPLER_PASSPHRASE")
+		if passphrase != "" {
+			utils.LogDebug("DOPPLER_PASSPHRASE setting passphrase")
```
Piccirello:
```suggestion
			utils.LogDebug("Setting passphrase via DOPPLER_PASSPHRASE env var")
```

Piccirello (commented):
Love the new tests

pkg/cmd/run.go
Piccirello: Just realized we have a `valueFromEnvironmentNotice` helper that you can use here

watsonian: Ah, perfect!

tests/e2e/run-fallback.sh
watsonian: @Piccirello I noticed this too when I was creating it. I assumed the incrementing number was to help you see which part of the test was failing easier. Was I not right there?

Piccirello: I didn't realize this was an existing pattern. I believe this was added during local testing and never cleaned up. It does seem like it could help identify the failing line, though I'd probably put it at the end to avoid printing an incorrect flag name (i.e. `ERROR: --passphrase flag is not respected (2)`). Either way not a huge deal.

Piccirello:
@watsonian you'll want to either squash your second commit or prepend `chore:` to it to avoid it appearing in the changelog. Also for your first commit you may want to rename to `Allow the passphrase` to `Allow fallback passphrase` so that it's more clear in the changelog what it's referring to.

watsonian:
Okay, I think we're good to go once the tests all pass.

Piccirello (approved):
🎉

PR #370, opened 2023-02-23
Log valueFromEnvironmentNotice messages to stderr

watsonian:
We had been printing these messages to `stdout`. This is fine for some operations, but others are more likely to be used in a scripted environment where just the value (e.g., `token=$(doppler configure get token --plain)`) is expected. By not printing to `stderr`, the message would end up in the value of the assignment. This just switches all uses of `valueFromEnvironmentNotice` to use `utils.Log` rather than `utils.Print`.


Piccirello (commented):
Could you update the commit message to "Log environment variable notice messages to stderr"? That'll make it more clear in the changelog.

watsonian:
Okay, should be good to go now!

Piccirello (approved):

PR #374, opened 2023-03-15
Add support for monorepo-friendly structuring in doppler.yaml setup file

watsonian:
This adds support for a more monorepo-friendly setup file. Historically, a setup file (`doppler.yaml`) had this format:

```
setup:
  project: cli
  config: dev
```

It could only contain values for a single project+config combination. If you had a monorepo that contained many services as subdirectories, you had to essentially add one of these to each of the subdirectories and then have a script go through running `doppler setup` in each directory (or clever usage of `find` on linux systems).

This change now supports setup files in this format:

```
setup:
  - project: cli
    config: dev
    path: .
  - project: example
    config: stg
    path: example/
```

It lets you specify multiple project+repo combinations and adds a new `path` field. If no `path` field is specified, then it refers to the root directory. If one is specified, it's relative to the root directory.

The change maintains backwards compatibility with old setup files while allowing the new usage. If the new format is detected, it will loop through, prompting you if you want to perform the setup for each directory (and prints the directory it's performing the setup for). `--no-interactive` will work as expected in this case. If the old format is detected, then everything works as it has historically.

Fixes ENG-3644.

### DIFF
--- a/pkg/cmd/setup.go
+++ b/pkg/cmd/setup.go
@@ -18,6 +18,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strings"
 
 	"github.com/DopplerHQ/cli/pkg/configuration"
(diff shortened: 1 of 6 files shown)


pkg/models/repo_config.go
```diff
@@ -16,10 +16,21 @@ limitations under the License.
 
 package models
 
-// RepoConfig holds all repo configuration
+// Config struct represents the basic project setup values
+type Config struct {
```
Piccirello: `Config` might be too generic of a name here since it'll be unique among all of our `pkg/models`.

pkg/cmd/setup.go
```diff
@@ -63,106 +64,115 @@ func setup(cmd *cobra.Command, args []string) {
...
+			// ignore when repo config is blank
+			(repo.Project == "" && repo.Config == "") ||
+				// ignore when project and config are already specified
+				(localConfig.EnclaveProject.Source == models.FlagSource.String() && localConfig.EnclaveConfig.Source == models.FlagSource.String())
+
+		// default to true so repo config is used on --no-interactive
+		useRepoConfig := true
+		if !ignoreRepoConfig && canPromptUser {
+			if len(repoConfig.Setup) > 1 && repo.Path != "" {
+				useRepoConfig = utils.ConfirmationPrompt(fmt.Sprintf("Use settings from repo config file (doppler.yaml) for %s?", expandedPath), true)
```
Piccirello, line 84: Wondering how we should handle the case where the user incorrectly specifies a doppler.yaml with multiple values but no path. Like this:

```yaml
setup:
  - project: cli
    config: dev
  - project: example
    config: stg
```

Should we error when `len(repoConfig.Setup) > 1 && repo.Path == ""`?

Piccirello (commented):
Did some quick testing locally and this works great!

pkg/models/repo_config.go
watsonian: Hm.. Yeah. I changed it to `ProjectConfig`.

pkg/cmd/setup.go
watsonian: Added an error for this!

pkg/cmd/setup.go
```diff
@@ -63,106 +64,118 @@ func setup(cmd *cobra.Command, args []string) {
...
+			utils.HandleError(errors.New("a path must be specified for all repos when more than one exists in the repo config file (doppler.yaml)"))
 		}
-
-		projects, httpErr := http.GetProjects(localConfig.APIHost.Value, utils.GetBool(localConfig.VerifyTLS.Value, true), localConfig.Token.Value, 1, 100)
-		if !httpErr.IsNil() {
-			utils.HandleError(httpErr.Unwrap(), httpErr.Message)
-		}
-		if len(projects) == 0 {
-			utils.HandleError(errors.New("you do not have access to any projects"))
+		expandedPath, _ := filepath.Abs(repo.Path)
```
Piccirello, line 71: What does `filepath.Abs()` return when `repo.Path` is an empty string?

Piccirello (dismissed):
LGTM! Requesting review from @nmanoogian as a sanity check.

pkg/cmd/setup.go
watsonian: From the docs:
>  If the path is empty, Base returns "."

watsonian: Oh wait, that's for a different function. Pretty sure it works the same. Let me double-check though.

watsonian: Confirmed – it uses the current working directory.

watsonian: https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/path/filepath/path.go;l=304-308
https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/path/filepath/path_windows.go;l=210-215

nmanoogian, line 71: This does appear to work for absolute URLs as well as relative ones. I think this is OK but I wanted to call it out:

```yaml
setup:
  - project: cli
    config: dev
    path: /Users/nic/Desktop
```

pkg/cmd/setup.go
```diff
@@ -63,106 +64,118 @@ func setup(cmd *cobra.Command, args []string) {
...
-		selectedProject = localConfig.EnclaveProject.Value
-	case models.EnvironmentSource.String():
-		utils.Log(valueFromEnvironmentNotice("DOPPLER_PROJECT"))
-		selectedProject = localConfig.EnclaveProject.Value
-	default:
-		if useRepoConfig && repoConfig.Setup.Project != "" {
-			utils.Print("Auto-selecting project from repo config file")
-			selectedProject = repoConfig.Setup.Project
-			break
+	for _, repo := range repoConfig.Setup {
```
nmanoogian, line 70: optional: If two setup structs contain the same path, the CLI will prompt for both and the second one will overwrite the first one:

```yaml
setup:
  - project: cli
    config: dev
    path: .
  - project: cli
    config: stg
    path: .
```

This is a nonsense configuration but we might want to consider adding a top-level check to prevent mistakes.

nmanoogian (dismissed):
Nice work on this! I did some light local testing; just two small callouts

pkg/cmd/setup.go
watsonian: I think that should be fine. The main concern was that it work for relative paths, which it does. If someone wants to use an absolute path there, I think that should be fine.

pkg/cmd/setup.go
watsonian: I'll take a look at this to see what we can do.

watsonian: Just pushed up a change that addresses this.

nmanoogian (dismissed):

watsonian:
@nmanoogian @Piccirello Tweaked the error checking a bit. Would love a second look at that when you have a chance.

nmanoogian (approved):

nmanoogian:
Nice refactor!

jieey1140:
so nice! I'm looking forward to this update :)

pkg/cmd/setup.go
```diff
@@ -237,6 +250,33 @@ func valueFromEnvironmentNotice(name string) string {
 	return fmt.Sprintf("Using %s from the environment. To disable this, use --no-read-env.", name)
 }
 
+// we're looking for duplicate paths and more than one repo being defined without a path.
+func setupFileErrorCheck(repos []models.ProjectConfig) {
+	// check to see if a repo isn't specifying a path and more than one repo exists
+	pathCount := make(map[string]int)
```
Piccirello, line 256: Should we resolve the absolute path before performing this check?

watsonian: Hm.. Not sure. In theory that shouldn't really matter in the end (duplicates would resolve in the same way). If we did resolve them first and then printed the path in the error, that path would also no longer actually match what they had in the `doppler.yaml` file, so it might even end up being confusing (e.g., `.` becomes `/home/foo/something`).

Piccirello: That fair, and we can always change this later if we want to revisit.

Piccirello (approved):

Piccirello:
Bypassing branch protection to merge this, despite the Salus failure. The failure is unrelated to our code: https://github.com/coinbase/salus/issues/828

ZeldOcarina:
Hi!
Can you please explain a bit how to run doppler commands using this?

watsonian:
@ZeldOcarina All Doppler CLI commands will remain unchanged! All this does is make it so you can create a single `doppler.yaml` file in the root of your monorepo directory and then run `doppler setup` once to configure all of the services in the repo in one command (previously, you would have needed a separate `doppler.yaml` file in each of your service's top-level subdirectory and would have had to run `doppler setup` in each directory one-by-one).

Once the setup is done, you won't need to specify the project or config in your CLI command because that's already setup.

ZeldOcarina:
Thank you!! I made it work somehow! It also helped me to handle this question:
https://stackoverflow.com/questions/76274391/how-do-i-allow-separate-env-variables-on-a-monorepo-for-each-project-in-vercel

PR #383, opened 2023-04-20
Add support for secret visibility types

apazzolini:
This PR adds CLI support for secret visibility types. It:
  - Correctly handles NULLs and includes the vistypes when using `doppler secrets --json`
  - Prints `[RESTRICTED]` in the table when appropriate
  - Adds the `--visibility` flag to explicitly show the visibility types in the table

Note that this should only be merged after visibility types has been released as the code expects the new properties from the API.

```
» doppler secrets
┌──────────────────────┬─────────────────────────────────────┬──────┐
│ NAME                 │ VALUE                               │ NOTE │
├──────────────────────┼─────────────────────────────────────┼──────┤
│ S1                   │ [RESTRICTED]                        │      │
│ S2                   │ v2                                  │      │
│ S3                   │ v3                                  │      │
└──────────────────────┴─────────────────────────────────────┴──────┘

» doppler secrets --visibility
┌──────────────────────┬────────────┬─────────────────────────────────────┬──────┐
│ NAME                 │ VISIBILITY │ VALUE                               │ NOTE │
├──────────────────────┼────────────┼─────────────────────────────────────┼──────┤
│ S1                   │ restricted │ [RESTRICTED]                        │      │
│ S2                   │ masked     │ v2                                  │      │
│ S3                   │ masked     │ v3                                  │      │
└──────────────────────┴────────────┴─────────────────────────────────────┴──────┘

» doppler secrets --visibility --raw
┌──────────────────────┬────────────┬──────────────┬────────────────┬──────────────────────────┬──────┐
│ NAME                 │ VISIBILITY │ VALUE        │ RAW VISIBILITY │ RAW VALUE                │ NOTE │
├──────────────────────┼────────────┼──────────────┼────────────────┼──────────────────────────┼──────┤
│ S1                   │ restricted │ [RESTRICTED] │ unmasked       │ ${stg.RESTRICTED_SECRET} │      │
│ S2                   │ masked     │ v2           │ masked         │ v2                       │      │
│ S3                   │ masked     │ v3           │ masked         │ v3                       │      │
└──────────────────────┴────────────┴──────────────┴────────────────┴──────────────────────────┴──────┘

» doppler secrets --json --raw | jq
{
  "S1": {
    "computed": null,
    "computedVisibility": "restricted",
    "note": "",
    "raw": "${stg.RESTRICTED_SECRET}",
    "rawVisibility": "unmasked"
  },
  "S2": {
    "computed": "v2",
    "computedVisibility": "masked",
    "note": "",
    "raw": "v2",
    "rawVisibility": "masked"
  },
  "S3": {
    "computed": "v3",
    "computedVisibility": "masked",
    "note": "",
    "raw": "v3",
    "rawVisibility": "masked"
  }
}
```

Closes ENG-6140

### DIFF
--- a/pkg/cmd/enclave_secrets.go
+++ b/pkg/cmd/enclave_secrets.go
@@ -97,6 +97,7 @@ func init() {
 	enclaveSecretsCmd.Flags().StringP("project", "p", "", "enclave project (e.g. backend)")
 	enclaveSecretsCmd.Flags().StringP("config", "c", "", "enclave config (e.g. dev)")
 	enclaveSecretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	enclaveSecretsCmd.Flags().Bool("visibility", false, "include secret visibility in table output")
 	enclaveSecretsCmd.Flags().Bool("only-names", false, "only print the secret names; omit all values")
 
 	enclaveSecretsGetCmd.Flags().StringP("project", "p", "", "enclave project (e.g. backend)")
@@ -104,6 +105,7 @@ func init() {
 	enclaveSecretsGetCmd.Flags().Bool("plain", false, "print values without formatting")
 	enclaveSecretsGetCmd.Flags().Bool("copy", false, "copy the value(s) to your clipboard")
 	enclaveSecretsGetCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	enclaveSecretsGetCmd.Flags().Bool("visibility", false, "include secret visibility in table output")
 	enclaveSecretsGetCmd.Flags().Bool("no-exit-on-missing-secret", false, "do not exit if unable to find a requested secret")
 	enclaveSecretsCmd.AddCommand(enclaveSecretsGetCmd)
 
--- a/pkg/cmd/secrets.go
+++ b/pkg/cmd/secrets.go
@@ -159,6 +159,7 @@ JSON Secret: "{\"logging\": \"info\"}"`,
 func secrets(cmd *cobra.Command, args []string) {
 	jsonFlag := utils.OutputJSON
 	raw := utils.GetBoolFlag(cmd, "raw")
+	visibility := utils.GetBoolFlag(cmd, "visibility")
 	onlyNames := utils.GetBoolFlag(cmd, "only-names")
 	localConfig := configuration.LocalConfig(cmd)
 
@@ -181,7 +182,7 @@ func secrets(cmd *cobra.Command, args []string) {
 			utils.HandleError(parseErr, "Unable to parse API response")
 		}
 
-		printer.Secrets(secrets, []string{}, jsonFlag, false, raw, false)
+		printer.Secrets(secrets, []string{}, jsonFlag, false, raw, false, visibility)
 	}
 }
 
@@ -190,6 +191,7 @@ func getSecrets(cmd *cobra.Command, args []string) {
 	plain := utils.GetBoolFlag(cmd, "plain")
 	copy := utils.GetBoolFlag(cmd, "copy")
 	raw := utils.GetBoolFlag(cmd, "raw")
+	visibility := utils.GetBoolFlag(cmd, "visibility")
 	exitOnMissingSecret := !utils.GetBoolFlag(cmd, "no-exit-on-missing-secret")
 	localConfig := configuration.LocalConfig(cmd)
 
@@ -226,7 +228,7 @@ func getSecrets(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	printer.Secrets(secrets, args, jsonFlag, plain, raw, copy)
+	printer.Secrets(secrets, args, jsonFlag, plain, raw, copy, visibility)
 }
 
 func setSecrets(cmd *cobra.Command, args []string) {
@@ -331,7 +333,7 @@ func setSecrets(cmd *cobra.Command, args []string) {
 	}
 
 	if !utils.Silent {
-		printer.Secrets(response, keys, jsonFlag, false, raw, false)
+		printer.Secrets(response, keys, jsonFlag, false, raw, false, false)
 	}
 }
 
@@ -363,7 +365,7 @@ func uploadSecrets(cmd *cobra.Command, args []string) {
 	}
 
 	if !utils.Silent {
-		printer.Secrets(response, []string{}, jsonFlag, false, raw, false)
+		printer.Secrets(response, []string{}, jsonFlag, false, raw, false, false)
 	}
 }
 
@@ -387,7 +389,7 @@ func deleteSecrets(cmd *cobra.Command, args []string) {
 		}
 
 		if !utils.Silent {
-			printer.Secrets(response, []string{}, jsonFlag, false, raw, false)
+			printer.Secrets(response, []string{}, jsonFlag, false, raw, false, false)
 		}
 	}
 }
@@ -590,6 +592,7 @@ func init() {
 	secretsCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
 	secretsCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	secretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	secretsCmd.Flags().Bool("visibility", false, "include secret visibility in table output")
 	secretsCmd.Flags().Bool("only-names", false, "only print the secret names; omit all values")
 
 	secretsGetCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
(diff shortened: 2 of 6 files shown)


pkg/printer/enclave.go
```diff
@@ -214,7 +214,11 @@ func Secrets(secrets map[string]models.ComputedSecret, secretsToPrint []string,
 		vals := []string{}
 		for _, name := range secretsToPrint {
 			if secrets[name] != (models.ComputedSecret{}) {
-				vals = append(vals, secrets[name].ComputedValue)
+				if secrets[name].ComputedValue != nil {
+					vals = append(vals, *secrets[name].ComputedValue)
+				} else {
+					vals = append(vals, "")
```
dp-franklin: Setting the clipboard to an empty string is reasonable but returning an error seems like the more correct thing to do here.

Supernit: Also we should consider moving the `copy` mechanism to the `getSecrets` func since that's the only thing using it.

dp-franklin (dismissed):
LGTM! I also did some testing.

pkg/models/parse.go
```diff
@@ -231,40 +231,30 @@ func ParseActivityLog(log map[string]interface{}) ActivityLog {
...
+	computed := map[string]ComputedSecret{}
+	for key, secret := range apiSecrets {
+		computed[key] = ComputedSecret{
+			Name:            key,
+			RawValue:        secret.RawValue,
+			ComputedValue:   secret.ComputedValue,
+			RawVisType:      secret.RawVisType,
+			ComputedVisType: secret.ComputedVisType,
+			Note:            secret.Note,
+		}
```
Piccirello: Will all of these values (like `note`) always be defined in the response?

pkg/cmd/secrets.go
```diff
@@ -159,6 +159,7 @@ JSON Secret: "{\"logging\": \"info\"}"`,
 func secrets(cmd *cobra.Command, args []string) {
 	jsonFlag := utils.OutputJSON
 	raw := utils.GetBoolFlag(cmd, "raw")
+	vistypes := utils.GetBoolFlag(cmd, "vistypes")
```
Piccirello: Torn on whether to call this `--vistypes` or `--visibility`

pkg/cmd/secrets.go
```diff
@@ -590,13 +592,15 @@ func init() {
...
 	secretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	secretsCmd.Flags().Bool("vistypes", false, "when printing the table, include secret visibility types")
 	secretsCmd.Flags().Bool("only-names", false, "only print the secret names; omit all values")
 
 	secretsGetCmd.Flags().StringP("project", "p", "", "project (e.g. backend)")
 	secretsGetCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	secretsGetCmd.Flags().Bool("plain", false, "print values without formatting")
 	secretsGetCmd.Flags().Bool("copy", false, "copy the value(s) to your clipboard")
 	secretsGetCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	secretsGetCmd.Flags().Bool("vistypes", false, "when printing the table, include secret visibility types")
```
Piccirello: Unfortunately you'll also need to add these flags to our legacy `doppler enclave secrets` command (`enclave_secrets.go`) or they'll error

pkg/printer/enclave.go
```diff
@@ -286,8 +286,15 @@ func Secrets(secrets map[string]models.ComputedSecret, secretsToPrint []string,
 		return
 	}
 
-	headers := []string{"name", "value"}
+	headers := []string{"name"}
+	if vistypes {
+		headers = append(headers, "vistype")
```
Piccirello: Here too, torn on whether to call this "vistype" or "visibility"

pkg/models/parse.go
apazzolini: Yes, all of these props are always included in the object version of the endpoint.

pkg/cmd/secrets.go
apazzolini: I struggled with this too. I settled on `vistypes` for consistency (e.g. the API deals with `rawVisType`, `computedVisType` for gets, `visType`, `originalVisType` for posts) and brevity. I think this is the right call; there's nothing else `visType` could be conflated with.

pkg/cmd/secrets.go
apazzolini: Ah, that's why the test was failing. Will add, thanks!

pkg/printer/enclave.go
apazzolini: :nod: - same as above here for me too, but brevity is arguably even more important here since `Raw Visibility` is a lot of characters.

pkg/cmd/secrets.go
Piccirello: Are you sold on `visType` vs switching to `visibility` throughout our API/CLI/etc? `visType` isn't much shorter than `visibility`, and I think `visibility` is more self-documenting than `visType`. `visType` makes more sense to me as an internal name, but not as a public-facing name.

apazzolini: I'm very much on the fence, but I can get behind `visibility` for the public-facing name. Will make it so.

dp-franklin (dismissed):
I re-tested the `secrets` and `secrets get` commands. Everything checks out.

pkg/models/api.go
```diff
@@ -17,10 +17,12 @@ package models
...
 type ComputedSecret struct {
-	Name          string `json:"name"`
-	RawValue      string `json:"raw"`
-	ComputedValue string `json:"computed"`
-	Note          string `json:"note"`
+	Name               string  `json:"name"`
+	RawValue           *string `json:"raw"`
+	ComputedValue      *string `json:"computed"`
+	RawVisibility      string  `json:"rawVisibility"`
+	ComputedVisibility string  `json:"computedVisibility"`
```
Piccirello, line 24: It's been a while so I don't remember - when will the raw and computed visibilities diverge?

pkg/printer/enclave.go
```diff
@@ -214,7 +214,11 @@ func Secrets(secrets map[string]models.ComputedSecret, secretsToPrint []string,
 		vals := []string{}
 		for _, name := range secretsToPrint {
 			if secrets[name] != (models.ComputedSecret{}) {
-				vals = append(vals, secrets[name].ComputedValue)
+				if secrets[name].ComputedValue == nil {
+					utils.HandleError(fmt.Errorf("Unable to copy to clipboard when value is restricted"))
```
Piccirello:
```suggestion
					utils.HandleError(fmt.Errorf("Unable to copy restricted value to clipboard"))
```

pkg/cmd/enclave_secrets.go
```diff
@@ -97,13 +97,15 @@ func init() {
 	enclaveSecretsCmd.Flags().StringP("project", "p", "", "enclave project (e.g. backend)")
 	enclaveSecretsCmd.Flags().StringP("config", "c", "", "enclave config (e.g. dev)")
 	enclaveSecretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
+	enclaveSecretsCmd.Flags().Bool("visibility", false, "when printing the table, include secret visibility types")
```
Piccirello:
```suggestion
	enclaveSecretsCmd.Flags().Bool("visibility", false, "include secret visibility in output")
```

Piccirello: To confirm, will the JSON output always include visibility info, or only when the `--visibility` flag is specified? We currently only include the `raw` field in our JSON output when `--raw` is specified.

pkg/models/api.go
apazzolini: If you have a non-restricted secret that references a restricted secret. (There's an example of this in the tables in the PR description if you want to see a real world example)

Piccirello: Ah, right. Ty.

pkg/cmd/enclave_secrets.go
apazzolini: The JSON output will always include the visibility for computed, and will include visibility for raw when `--raw` is specified (which is why my prior wording on the visibility flag was narrowed to the table output). `--visibility` has no impact on the JSON output.

I opted for this because we're now potentially returning `null` in the JSON output for restricted secrets, and I wanted the visibility to be present to explain that behavior. Do you disagree?

Piccirello: That's tough, I could go either way. I think your point is valid and can get behind your current implementation.

apazzolini: 👍 in light of the flag only impacting the table output, I reworded to "include secret visibility in table output"

Piccirello (approved):

PR #386, opened 2023-04-21
ENG-6137: Fix bug where interactive setups weren't working

watsonian:
This fixes a bug introduced in #374 that caused interactive invocations of `doppler setup` to operate as no-ops.

### DIFF
--- a/pkg/controllers/repo_config.go
+++ b/pkg/controllers/repo_config.go
@@ -68,6 +68,12 @@ func RepoConfig() (models.MultiRepoConfig, Error) {
 		return repoConfig, Error{}
 	} else if utils.Exists(ymlRepoConfigFile) {
 		utils.LogWarning(fmt.Sprintf("Found %s file, please rename to %s for repo configuration", ymlRepoConfigFile, repoConfigFileName))
+	} else {
+		// If no config file exists, then this is for an interactive setup, so
+		// return a MultiRepoConfig object containing an empty ProjectConfig object
+		var repoConfig models.MultiRepoConfig
+		repoConfig.Setup = append(repoConfig.Setup, models.ProjectConfig{})
+		return repoConfig, Error{}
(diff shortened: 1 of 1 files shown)


pkg/controllers/repo_config.go
```diff
@@ -68,6 +68,12 @@ func RepoConfig() (models.MultiRepoConfig, Error) {
 		return repoConfig, Error{}
 	} else if utils.Exists(ymlRepoConfigFile) {
 		utils.LogWarning(fmt.Sprintf("Found %s file, please rename to %s for repo configuration", ymlRepoConfigFile, repoConfigFileName))
+	} else {
+		// If no config file exists, then this is for an interactive setup, so
+		// return a MultiRepoConfig object containing an empty ProjectConfig object
+		var repoConfig models.MultiRepoConfig
+		repoConfig.Setup = append(repoConfig.Setup, models.ProjectConfig{})
+		return repoConfig, Error{}
```
mrasnake-doppler, line 76: Just a note, but from a Go standards perspective I don't think I like what we are doing with this `Error` object. But I get that it would require a greater refactor then the scope of this issue. Something I will look into when I have cycles for a redesign. 

Other then that I will approve the Go code, but recommend you also wait for an approval from someone with a deeper understanding of the business logic.

mrasnake-doppler (approved):
Just a note, but from a Go standards perspective I don't think I like what we are doing with this Error object. But I get that it would require a greater refactor then the scope of this issue. Something I will look into when I have cycles for a redesign.

Other then that I will approve the Go code, but recommend you also wait for an approval from someone with a deeper understanding of the business logic.

watsonian:
We've decided to roll back to prior to this change so we have more time to review things to make sure there aren't any other issues surrounding this. I'm going to close this PR for now.

PR #389, opened 2023-04-25
Fix salus failures

nmanoogian:
We started to notice errors from Salus relating to `cargo-audit`, which we don't use, so we now define the specific scanners for Salus to run.

The GitHub Action that we're using automatically pulls the latest Salus image which included some semgrep changes that seem to break with our custom rules. For this reason, I've also pulls semgrep into a separate step.

### DIFF
--- a/pkg/http/http.go
+++ b/pkg/http/http.go
@@ -226,7 +226,11 @@ func performRequest(req *http.Request, verifyTLS bool) (int, http.Header, []byte
 		resp, err := client.Do(req) // nosemgrep: trailofbits.go.invalid-usage-of-modified-variable.invalid-usage-of-modified-variable
 		if err != nil {
 			if resp != nil {
-				defer resp.Body.Close()
+				defer func() {
+					if closeErr := resp.Body.Close(); closeErr != nil {
+						utils.LogDebug(closeErr.Error())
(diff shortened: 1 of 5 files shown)


pkg/http/http.go
```diff
@@ -226,7 +226,7 @@ func performRequest(req *http.Request, verifyTLS bool) (int, http.Header, []byte
 		resp, err := client.Do(req) // nosemgrep: trailofbits.go.invalid-usage-of-modified-variable.invalid-usage-of-modified-variable
 		if err != nil {
 			if resp != nil {
-				defer resp.Body.Close()
+				defer resp.Body.Close() // #nosec: G307
 			}
```
Piccirello: I think ignoring this error is fine given that this is an HTTP request and not a file. Wanted to share this snippet in case we do want to handle this

https://github.com/securego/gosec/issues/512#issuecomment-675286833

Piccirello (dismissed):

pkg/http/http.go
nmanoogian: Great find 👍 I'd definitely prefer resolving the error, rather than ignoring it. Just push the change

Piccirello (approved):

PR #325, opened 2022-08-17
Add install support for Git Bash for Windows

watsonian:
This adds support for Git Bash for Windows to the install script.

Solves #302.

PR #391, opened 2023-05-01
Allow MINGW64 environments to update

watsonian:
We added support for installing the CLI via the shell script installation method under Git Bash in #325. That works great, but when you try updating via `doppler update` after doing that, you're greeted with a `this command is not yet implemented for your operating system` error. This update adds an additional check that looks for a value of `MINGW64` in the `MSYSTEM` environment variable (which Git Bash sets automatically) and if it finds that allows for updates (the typical Go OS checks return a generic "windows" for Git Bash, which is why we had to resort to the environment variable). It also swaps to executing the install script inside of `sh`.

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -64,7 +64,15 @@ func RunInstallScript() (bool, string, Error) {
 	command := []string{tmpFile, "--debug"}
 
 	startTime = time.Now()
-	out, err := exec.Command(command[0], command[1:]...).CombinedOutput() // #nosec G204
+	var out []byte
+	if utils.IsWindows() {
+		// executing in sh on Windows avoids errors like this:
+		// Doppler Error: fork/exec C:\...\.install.sh.1063970983: %1 is not a valid Win32 application.
+		out, err = exec.Command("sh", command...).CombinedOutput() // #nosec G204
+	} else {
+		out, err = exec.Command(command[0], command[1:]...).CombinedOutput() // #nosec G204
+	}
+
 	executeDuration := time.Now().Sub(startTime).Milliseconds()
 
 	strOut := string(out)
--- a/pkg/cmd/update.go
+++ b/pkg/cmd/update.go
@@ -30,7 +30,7 @@ var updateCmd = &cobra.Command{
 	Short: "update the Doppler CLI",
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, args []string) {
-		if utils.IsWindows() {
+		if !utils.CanUpdate() {
 			utils.HandleError(fmt.Errorf("this command is not yet implemented for your operating system"))
 		}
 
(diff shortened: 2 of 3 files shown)


pkg/controllers/update.go
```diff
@@ -64,7 +64,7 @@ func RunInstallScript() (bool, string, Error) {
 	command := []string{tmpFile, "--debug"}
 
 	startTime = time.Now()
-	out, err := exec.Command(command[0], command[1:]...).CombinedOutput() // #nosec G204
+	out, err := exec.Command("sh", command...).CombinedOutput() // #nosec G204
```
Piccirello: Could you pull this into a separate commit? Also, could you provide some more info on why this changed is needed?

watsonian: Yep! The change was due to this happening in Git Bash when it attempted to execute the shell script directly rather than inside `sh`:

```
$ ./cli.exe update
Updating...
Unable to install the latest Doppler CLI
Doppler Error: fork/exec C:\Users\watsonian\.install.sh.1063970983: %1 is not a valid Win32 application.
```

Executing inside `sh` avoids this issue and works more or less identically to how it did before. I could swap it to executing `/usr/bin/env` with `sh` as the first parameter too though. That would be more consistent with how it was working previously.

watsonian: Actually, that doesn't seem to work either even though it's definitely there... 🤔 

```
watsonian@JOELWATSON638E MINGW64 /home/watsonian/dev
$ ./cli.exe update
Updating...
Unable to install the latest Doppler CLI
Doppler Error: exec: "/usr/bin/env": file does not exist

watsonian@JOELWATSON638E MINGW64 /home/watsonian/dev
$ which env
/usr/bin/env
```

Doesn't seem to work properly with just `env` either. Maybe there's another more windows-friendly way of executing the command?

watsonian: I take that back. Using just `env` does work as expected. I had forgotten that I'd not already installed the CLI in my fresh Windows 11 install, so it was erroring out with a 127 since it wasn't finding something to update (the dev build of the CLI wasn't on the PATH).

Do we want to go that route? It seems unlikely that `sh` wouldn't be on the PATH already if its installed on a system (and it definitely is in Git Bash).

Piccirello: This is difficult to say without testing this command thoroughly. I'm wondering if we can only run this alternative command when on Windows?

watsonian: We could probably do that easily enough. Shall I go that route?

watsonian: Latest force-push addresses this. [compare](https://github.com/DopplerHQ/cli/compare/71bc30575f0e8e23b3bf83bcdb10b5f70a3c6eb3..4d904c7c2777173359da987e4385909e5487117c)

nmanoogian (approved):

Piccirello (approved):
This is a nice improvement!

PR #393, opened 2023-05-08
Fix projects delete regression

rgharris:
Introduced in #354. Other similar issues were fixed in #366. I don't think there are any further incidents like this after manual review.

Closes ENG-6239


Piccirello (approved):
#354 strikes again. I did another manual review after #366 but still missed this...

PR #394, opened 2023-05-10
Add support for monorepo-friendly doppler.yaml files (take 2)

watsonian:
Take 2 of #374 which we had to revert due to a bug that broke interactive invocations of `doppler setup`. This reintroduces the same change in #374 with the addition of the fix in #386 that we decided to hold off on in favor of rolling back so we could do more testing. It also adds an additional test for interactive `doppler setup` invocations using `expect`. I've run the test without the fix from #386 to ensure it would have caught it (which it would have).

-----

This adds support for a more monorepo-friendly setup file. Historically, a setup file (`doppler.yaml`) had this format:

```
setup:
  project: cli
  config: dev
```

It could only contain values for a single project+config combination. If you had a monorepo that contained many services as subdirectories, you had to essentially add one of these to each of the subdirectories and then have a script go through running `doppler setup` in each directory (or clever usage of `find` on linux systems).

This change now supports setup files in this format:

```
setup:
  - project: cli
    config: dev
    path: .
  - project: example
    config: stg
    path: example/
```

It lets you specify multiple project+repo combinations and adds a new `path` field. If no `path` field is specified, then it refers to the root directory. If one is specified, it's relative to the root directory.

The change maintains backwards compatibility with old setup files while allowing the new usage. If the new format is detected, it will loop through, prompting you if you want to perform the setup for each directory (and prints the directory it's performing the setup for). `--no-interactive` will work as expected in this case. If the old format is detected, then everything works as it has historically.

Fixes ENG-3644.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,8 @@ jobs:
       with:
         go-version: '1.19'
         check-latest: true
+    - name: Install expect
+      run: sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq expect
     - name: Checkout
       uses: actions/checkout@v1
       with:
--- a/tests/e2e/setup.sh
+++ b/tests/e2e/setup.sh
@@ -1,0 +1,210 @@
+#!/bin/bash
+
+set -euo pipefail
+
+TEST_NAME="setup file"
+TEST_CONFIG_DIR="./temp-config-dir"
+DOPPLER_PROJECT=""
+DOPPLER_CONFIG=""
+
+cleanup() {
+  exit_code=$?
+  if [ "$exit_code" -ne 0 ]; then
+    echo "ERROR: '$TEST_NAME' tests failed during execution"
+    afterAll || echo "ERROR: Cleanup failed"
+  fi
+
+  exit "$exit_code"
+}
+trap cleanup EXIT
+
+beforeAll() {
+  echo "INFO: Executing '$TEST_NAME' tests"
+  mv doppler.yaml doppler.yaml.bak
+}
+
+beforeEach() {
+  header
+  rm -rf $TEST_CONFIG_DIR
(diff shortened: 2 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -11,6 +11,8 @@ jobs:
       with:
         go-version: '1.19'
         check-latest: true
+    - name: Install expect
+      run: sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq expect
```
Piccirello, line 15: Is `expect` needed given that our use of `doppler setup` won't actually be interactive? (due to the token only having access to a single project/config)

tests/e2e/setup.sh
```diff
@@ -71,6 +71,58 @@ beforeAll
 ######################################################################
 #
 
+name="interactive setup"
```
Piccirello, line 74: Can you prepend the `expect` commit with `chore:` so that it doesn't show up in the changelog?

Piccirello (commented):
For other reviewers: based on a [diff](https://github.com/DopplerHQ/cli/compare/3bc72d1b78afffdadfe89e02cfef1c6db11478c7..80683f9d9988f8d9d5c045e77f651cf2165c5da2), it appears the relevant change from the previous PR is the addition of the `else` condition in `pkg/controllers/repo_config.go`.

.github/workflows/test.yml
watsonian: Now that you mention it, we can probably just grep for the output. Do we want to keep the pattern there for future cases where we might want to use it so we have a reference?

Piccirello: Seems reasonable enough!

Piccirello (approved):

nmanoogian (approved):

PR #401, opened 2023-06-02
Support passing list of secrets to --preserve-env flag

Piccirello:
This PR changes the `--preserve-env` flag from a boolean flag to a string flag. It supports the same syntax as before (i.e. `--preserve-env` is equivalent to `--preserve-env=true`) while allowing you to specify a comma separated list of secrets to override. For example, `DEBUG=1 doppler run --preserve-env="DEBUG" -- printenv DEBUG`. To ensure backwards compatibility, I verified that the tests in the first commit passed before pushing up the second commit.

Closes ENG-6327.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -86,7 +86,7 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		fallbackReadonly := utils.GetBoolFlag(cmd, "fallback-readonly")
 		fallbackOnly := utils.GetBoolFlag(cmd, "fallback-only")
 		exitOnWriteFailure := !utils.GetBoolFlag(cmd, "no-exit-on-write-failure")
-		preserveEnv := utils.GetBoolFlag(cmd, "preserve-env")
+		preserveEnv := cmd.Flag("preserve-env").Value.String()
 		forwardSignals := utils.GetBoolFlag(cmd, "forward-signals")
 		localConfig := configuration.LocalConfig(cmd)
 		dynamicSecretsTTL := utils.GetDurationFlag(cmd, "dynamic-ttl")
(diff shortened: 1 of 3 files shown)


pkg/cmd/run.go
```diff
@@ -688,7 +694,11 @@ func init() {
 	runCmd.Flags().StringP("config", "c", "", "config (e.g. dev)")
 	runCmd.RegisterFlagCompletionFunc("config", configNamesValidArgs)
 	runCmd.Flags().String("command", "", "command to execute (e.g. \"echo hi\")")
-	runCmd.Flags().Bool("preserve-env", false, "ignore any Doppler secrets that are already defined in the environment. this has potential security implications, use at your own risk.")
+	// note: requires using "--preserve-env=VALUE", doesn't work with "--preserve-env VALUE"
```
watsonian, line 697: Is this a Cobra idiosyncrasy?

Piccirello: It is, afaik. Their [docs](https://github.com/spf13/pflag#command-line-flag-syntax) indicate that a space is only allowed on "non-boolean and flags without a 'no option default value'", and this flag now has a "no option default value".

watsonian: Well, that's annoying. Will this be reflected in the help output automatically? Or should we update the flag description?

erquhart: User that requested this here 👋 - I'd take having this critical functionality in it's current form over not having it. Possible to merge and iterate if/when necessary?

Piccirello: We're still waiting on reviews from the rest of the team but should have this shipped this week.

> Well, that's annoying. Will this be reflected in the help output automatically? Or should we update the flag description?

I'll add a note to the flag description.

nmanoogian (approved):

PR #404, opened 2023-06-12
Add lower kebab name transformer

watsonian:
This adds a new `lower-kebab` name transformer that will transform secrets from `SOME_SECRET` to `some-secret`.

Closes ENG-6375.

### DIFF
--- a/pkg/models/secrets_name_transformer.go
+++ b/pkg/models/secrets_name_transformer.go
@@ -31,6 +31,11 @@ var CamelTransformer = &SecretsNameTransformer{
 	Type:      "camel",
 	EnvCompat: true,
 }
+var LowerKebabTransformer = &SecretsNameTransformer{
+	Name:      "Lower Kebab",
+	Type:      "lower-kebab",
+	EnvCompat: true,
+}
 var LowerSnakeTransformer = &SecretsNameTransformer{
 	Name:      "Lower Snake",
 	Type:      "lower-snake",
@@ -55,6 +60,7 @@ var DotNETEnvTransformer = &SecretsNameTransformer{
 var SecretsNameTransformersList = []*SecretsNameTransformer{
 	UpperCamelTransformer,
 	CamelTransformer,
+	LowerKebabTransformer,
 	LowerSnakeTransformer,
 	TFVarTransformer,
 	DotNETTransformer,


pkg/utils/name_transformers.go
```diff
@@ -34,6 +34,20 @@ func UpperCamel(name string) string {
 	return strings.Join(parts, "")
 }
 
+func LowerKebab(name string) string {
```
nmanoogian: It doesn't look like this function or `UpperCamel` are actually used because name transformation is happening on the server. `DotNETNameTransform` (via `MapToDotNETJSONFormat`) and `MapToEnvFormat` are used with mount formats via `SecretsToBytes` but it looks like that's exclusively for custom formats so you shouldn't need to add one of those

nmanoogian (dismissed):

pkg/utils/name_transformers_test.go
```diff
@@ -42,6 +42,23 @@ func TestUpperCamel(t *testing.T) {
 	}
 }
 
+func TestLowerKebab(t *testing.T) {
+	testCases := []testCase{
+		{"TEST", "test"},
+		{"TEST_", "test"},
```
Piccirello: I wonder about this test case. Should we not return `test-` to more accurately match the user's input?

pkg/utils/name_transformers_test.go
```diff
@@ -42,6 +42,23 @@ func TestUpperCamel(t *testing.T) {
...
+		{"TEST_SECRET", "test-secret"},
+		{"TEST_SECRET_NAME", "test-secret-name"},
+		{"TEST__SECRET", "test-secret"},
```
Piccirello: Same here - should this generate `test--secret`?

watsonian: We do that server-side, but the UpperCamel transformer does the same thing:

https://github.com/DopplerHQ/cli/blob/master/pkg/utils/name_transformers_test.go#L34

pkg/utils/name_transformers_test.go
watsonian: This is basically just following suite with the UpperCamel and DotNET transformers. If we change this behavior here, we probably would want to change it for the other too as well.

pkg/utils/name_transformers.go
watsonian: Ah, okay. This will actually force the change @Piccirello was mentioning as well since we already have the behavior he mentions serverside.

pkg/utils/name_transformers_test.go
Piccirello: I believe `DotNET` does that on purpose, whereas I'm not sure there's a good reason UpperCamel does it. I think it's more correct to just convert all underscores to hyphens server-side.

Piccirello (approved):

nmanoogian (approved):

PR #408, opened 2023-06-16
Add support for restarting process when secrets change

Piccirello:
This PR adds a `--watch` flag to the `run` command that will automatically restart the user's process when secrets change. It monitors for changed secrets using a long-lived request. When secrets change, the CLI fetches the latest secrets, sends a SIGTERM to the process, waits up to 10 seconds, sends a SIGKILL if necessary, and then starts the process again.

This feature is currently in beta and is not yet broadly available in customer workplaces. Given this, the feature will not be mentioned in the CLI's release notes.

Closes ENG-6390

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -140,8 +140,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			exitOnWriteFailure: exitOnWriteFailure,
 			passphrase:         passphrase,
 		}
-		// retrieve secrets
-		dopplerSecrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL, format, secretsToInclude)
 
 		mountPath := cmd.Flag("mount").Value.String()
 		mountFormatString := cmd.Flag("mount-format").Value.String()
@@ -151,28 +149,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		shouldAutoDetectFormat := !cmd.Flags().Changed("mount-format")
 		shouldMountFile := mountPath != ""
 		shouldMountTemplate := mountTemplate != ""
-
-		// The potentially dangerous secret names only are only dangerous when they are set
-		// as environment variables since they have the potential to change the default shell behavior.
-		// When mounting the secrets into a file these are not dangerous
-		if !shouldMountFile {
-			if err := controllers.CheckForDangerousSecretNames(dopplerSecrets); err != nil {
-				utils.LogWarning(err.Error())
-			}
-		}
-
-		if len(secretsToInclude) > 0 {
-			noExitOnMissingIncludedSecrets := cmd.Flags().Changed("no-exit-on-missing-only-secrets")
-			missingSecrets := controllers.MissingSecrets(dopplerSecrets, secretsToInclude)
-			if len(missingSecrets) > 0 {
-				err := fmt.Errorf("the following secrets you are trying to include do not exist in your config:\n- %v", strings.Join(missingSecrets, "\n- "))
-				if noExitOnMissingIncludedSecrets {
-					utils.LogWarning(err.Error())
-				} else {
-					utils.HandleError(err)
-				}
-			}
-		}
 
 		var mountFormat string
 		if mountFormatVal, ok := models.SecretsMountFormatMap[mountFormatString]; ok {
@@ -203,13 +179,8 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			existingEnvKeys[key] = value
 		}
 
-		env := []string{}
-		secrets := map[string]string{}
-		var onExit func()
+		var templateBody string
 		if shouldMountFile {
-			secrets = dopplerSecrets
-			env = originalEnv
-
 			if shouldAutoDetectFormat {
 				if shouldMountTemplate {
 					mountFormat = models.TemplateMountFormat
@@ -232,7 +203,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 
 			utils.LogDebug(fmt.Sprintf("Using %s format", mountFormat))
 
-			var templateBody string
 			if shouldMountTemplate {
 				if mountFormat != models.TemplateMountFormat {
 					utils.HandleError(errors.New("--mount-template can only be used with --mount-format=template"))
@@ -241,6 +211,39 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			} else if mountFormat == models.TemplateMountFormat {
 				utils.HandleError(errors.New("--mount-template must be specified when using --mount-format=template"))
 			}
+		}
+
+		// retrieve secrets
+		dopplerSecrets := fetchSecrets(localConfig, enableCache, fallbackOpts, metadataPath, nameTransformer, dynamicSecretsTTL, format, secretsToInclude)
+
+		// The potentially dangerous secret names only are only dangerous when they are set
+		// as environment variables since they have the potential to change the default shell behavior.
+		// When mounting the secrets into a file these are not dangerous
+		if !shouldMountFile {
+			if err := controllers.CheckForDangerousSecretNames(dopplerSecrets); err != nil {
+				utils.LogWarning(err.Error())
+			}
+		}
+
+		if len(secretsToInclude) > 0 {
+			noExitOnMissingIncludedSecrets := cmd.Flags().Changed("no-exit-on-missing-only-secrets")
+			missingSecrets := controllers.MissingSecrets(dopplerSecrets, secretsToInclude)
+			if len(missingSecrets) > 0 {
+				err := fmt.Errorf("the following secrets you are trying to include do not exist in your config:\n- %v", strings.Join(missingSecrets, "\n- "))
+				if noExitOnMissingIncludedSecrets {
+					utils.LogWarning(err.Error())
+				} else {
+					utils.HandleError(err)
+				}
+			}
+		}
+
+		env := []string{}
+		secrets := map[string]string{}
+		var onExit func()
+		if shouldMountFile {
+			secrets = dopplerSecrets
+			env = originalEnv
 
 			secretsBytes, err := controllers.SecretsToBytes(secrets, mountFormat, templateBody)
 			if !err.IsNil() {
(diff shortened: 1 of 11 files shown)


pkg/controllers/secrets.go
```diff
@@ -159,17 +163,28 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 			// this operation blocks until the FIFO is opened for reads
 			f, err := os.OpenFile(mountPath, os.O_WRONLY, os.ModeNamedPipe) // #nosec G304
 			if err != nil {
+				if !fifoExists && errors.Is(err, fs.ErrNotExist) {
+					break
+				}
```
nmanoogian, line 217: check: Do we want to call `cleanupFIFO()` in this case as well?

pkg/controllers/secrets.go
```diff
@@ -127,8 +127,12 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 		return "", nil, Error{Err: err, Message: "Unable to mount secrets file"}
 	}
 
+	fifoExists := true
+
 	// cleanup named pipe on exit
 	cleanupFIFO := func() {
+		fifoExists = false
```
nmanoogian: nit: This var might be better named `fifoCleanupStarted` (with negated values) because we don't really know if it exists at this point

pkg/utils/retry.go
```diff
@@ -37,14 +37,18 @@ func retry(attempts int, sleep time.Duration, f func() error) error {
...
-			return retry(attempts, 2*sleep, f)
+			return Retry(attempts, 2*sleep, f)
 		}
 		return err
 	}
 
 	return nil
 }
 
+func StopRetryError(err error) StopRetry {
```
nmanoogian, line 48: check: Why not just let callers build the `StopRetry` struct directly?

pkg/controllers/secrets.go
```diff
@@ -60,6 +64,24 @@ var dangerousSecretNames = [...]string{
...
+	Enable             bool
+	Path               string
+	LegacyPath         string
+	Readonly           bool
+	Exclusive          bool
+	ExitOnWriteFailure bool
+	Passphrase         string
+}
+
+type MountOptions struct {
```
nmanoogian, line 77: Nice! Containing these options in structs makes them a lot easier to work with

pkg/controllers/secrets.go
```diff
@@ -273,3 +295,322 @@ func CheckForDangerousSecretNames(secrets map[string]string) error {
...
+	// as environment variables since they have the potential to change the default shell behavior.
+	// When mounting the secrets into a file these are not dangerous
+	if !mountOptions.Enable {
+		if err := CheckForDangerousSecretNames(secrets); err != nil {
+			utils.LogWarning(err.Error())
+		}
+	}
+}
+
+func PrepareSecrets(dopplerSecrets map[string]string, originalEnv []string, preserveEnv string, existingEnvKeys map[string]string, mountOptions MountOptions) ([]string, func()) {
```
nmanoogian: recommendation/optional: `existingEnvKeys` is derived from `originalEnv` and only used within this function; thoughts on moving that logic here?

```diff
diff --git a/pkg/cmd/run.go b/pkg/cmd/run.go
index fc6d3ae..a246057 100644
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -163,16 +163,6 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			utils.HandleError(errors.New("--mount-template must be used with --mount"))
 		}
 
-		originalEnv := os.Environ()
-		existingEnvKeys := map[string]string{}
-		for _, envVar := range originalEnv {
-			// key=value format
-			parts := strings.SplitN(envVar, "=", 2)
-			key := parts[0]
-			value := parts[1]
-			existingEnvKeys[key] = value
-		}
-
 		var templateBody string
 		if shouldMountFile {
 			if shouldAutoDetectFormat {
@@ -289,7 +279,7 @@ doppler run --mount secrets.json -- cat secrets.json`,
 			terminatedByWatch = false
 
 			var env []string
-			env, cleanupMount = controllers.PrepareSecrets(secrets, originalEnv, preserveEnv, existingEnvKeys, mountOptions)
+			env, cleanupMount = controllers.PrepareSecrets(secrets, os.Environ(), preserveEnv, mountOptions)
 
 			global.WaitGroup.Add(1)
 
diff --git a/pkg/controllers/secrets.go b/pkg/controllers/secrets.go
index c437bc4..62e4c6e 100644
--- a/pkg/controllers/secrets.go
+++ b/pkg/controllers/secrets.go
@@ -319,7 +319,19 @@ func ValidateSecrets(secrets map[string]string, secretsToInclude []string, exitO
 	}
 }
 
-func PrepareSecrets(dopplerSecrets map[string]string, originalEnv []string, preserveEnv string, existingEnvKeys map[string]string, mountOptions MountOptions) ([]string, func()) {
+func getEnvKeys(env []string) map[string]string {
+	existingEnvKeys := map[string]string{}
+	for _, envVar := range env {
+		// key=value format
+		parts := strings.SplitN(envVar, "=", 2)
+		key := parts[0]
+		value := parts[1]
+		existingEnvKeys[key] = value
+	}
+	return existingEnvKeys
+}
+
+func PrepareSecrets(dopplerSecrets map[string]string, originalEnv []string, preserveEnv string, mountOptions MountOptions) ([]string, func()) {
 	env := []string{}
 	secrets := map[string]string{}
 	var onExit func()
@@ -341,6 +353,8 @@ func PrepareSecrets(dopplerSecrets map[string]string, originalEnv []string, pres
 		// export path to mounted file
 		env = append(env, fmt.Sprintf("%s=%s", "DOPPLER_CLI_SECRETS_PATH", mountPath))
 	} else {
+		existingEnvKeys := getEnvKeys(originalEnv)
+
 		// remove any reserved keys from secrets
 		reservedKeys := []string{"PATH", "PS1", "HOME"}
 		for _, reservedKey := range reservedKeys {


```

pkg/http/api.go
```diff
@@ -156,6 +156,39 @@ func RevokeAuthToken(host string, verifyTLS bool, token string) (map[string]inte
 	return result, Error{}
 }
 
+// DownloadSecrets for specified project and config
```
nmanoogian: nit: copy-pasta?

pkg/cmd/run.go
```diff
@@ -213,113 +207,240 @@ doppler run --mount secrets.json -- cat secrets.json`,
...
+			if secretsFetchedAt.After(lastSecretsFetch) {
+				lastSecretsFetch = secretsFetchedAt
+			}
+
+			controllers.ValidateSecrets(secrets, secretsToInclude, exitOnMissingIncludedSecrets, mountOptions)
+
+			isRestart := c != nil
+			// terminate the old process
+			if isRestart {
+				terminatedByWatch = true
```
nmanoogian, line 239: This mutation of `terminatedByWatch` is not protected by a mutex, should it be?

pkg/cmd/run.go
```diff
@@ -241,82 +205,242 @@ doppler run --mount secrets.json -- cat secrets.json`,
...
+					os.Exit(exitCode)
 				}
-				// then use doppler secrets
-				for name, value := range dopplerSecrets {
-					secrets[name] = value
+			}()
+		}
+
+		watchHandler := func(data []byte) {
+			// Expected format: "event: message\ndata: {JSON}\n\n"
```
nmanoogian: recommendation/option: What do you think about pulling the event parsing into a helper func? I think it's making this function a little too heavy

nmanoogian (commented):
Looking great! I love the refactoring; much cleaner.

I haven't had the chance to test yet but will soon

pkg/utils/retry.go
Piccirello: 🤷

pkg/controllers/secrets.go
Piccirello: In this case we weren't necessarily the party that created the FIFO (since we never successfully opened it), so we don't want to delete it.

pkg/controllers/secrets.go
Piccirello: +1

pkg/cmd/run.go
Piccirello: Good call

pkg/cmd/run.go
Piccirello: I don't believe a race is possible here given our use of `watchMutex`. `watchMutex` ensures that `startProcess()` cannot be invoked concurrently except on the initial watch event. The initial call to `startProcess()` (before the initial watch event) would run with `c == nil`, so `terminatedByWatch` won't be mutated. Thus I think we're safe here.

In the worst case, a mutation to `terminatedByWatch` would mean us erroneously restarting an exited process rather than the CLI exiting.

pkg/controllers/secrets.go
Piccirello: On further review, `utils.CreateNamedPipe` would have already completed successfully above. I'll add a call to `cleanupFIFO()` here.

pkg/controllers/secrets.go
Piccirello: I've made this logic a little easier to follow by also reordering the conditions.

Piccirello:
All feedback [addressed](https://github.com/DopplerHQ/cli/compare/8914f0c668456404ef458f09d7260c1711e504b3..00f45d5898abd233c4edbe53e055a52bc05391fb).

pkg/cmd/run.go
nmanoogian: That makes sense; thanks for the explanation!

pkg/controllers/watch.go
```diff
@@ -0,0 +1,59 @@
...
+	if !strings.HasPrefix(parts[1], dataPrefix) {
+		utils.LogDebug("Unable to parse API response; invalid data")
+		return models.WatchSecrets{}
+	}
+
+	dataJson := strings.TrimPrefix(parts[1], dataPrefix)
+	var watchSecrets models.WatchSecrets
+	err := json.Unmarshal([]byte(dataJson), &watchSecrets)
+	if err != nil {
+		CaptureEvent("WatchDataParseError", nil)
```
nmanoogian: check: Do you want to be capturing these events in the other failure scenarios as well?

nmanoogian (dismissed):
Updates look solid 👍

pkg/controllers/watch.go
Piccirello: Definitely, good catch. Added in [this force push](https://github.com/DopplerHQ/cli/compare/f4eaf267a2c6a02e3fdfa373fc4e9b924f5daea3..44e7e08442ae4d21000f1d4112de8035be2889cf).

nmanoogian (approved):

PR #412, opened 2023-06-29
chore: all commands help text is now sentence case

ryan-blunden:
Fixed command help strings that started with a lowercase character.


Piccirello (approved):
Thanks Ryan!

PR #413, opened 2023-06-29
TUI Beta

apazzolini:
This PR implements a textual user interface for Doppler. It took heavy inspiration (both visually and in code) from [lazygit](https://github.com/jesseduffield/lazygit) and is launched via `doppler tui`.

It currently supports switching between projects and configs and allows you to filter for, view, edit, add, delete, and copy secrets. For now, the keybindings and colors are hardcoded, but we could add the ability for users to customize these in later versions. Pressing `?` will show a popup describing the available actions.

I've been using this for the last few months without issue, but this should be considered an **alpha** implementation until it's been used a bit more widely.

image
image


pkg/tui/common/logging.go
```diff
@@ -0,0 +1,52 @@
...
+	var log *logrus.Logger
+	if utils.DebugTUI {
+		log = newDebugLogger()
+	} else {
+		log = newDiscardedLogger()
+	}
+
+	// highly recommended: tail -f development.log | humanlog
+	// https://github.com/aybabtme/humanlog
+	// log.Formatter = &logrus.JSONFormatter{}
```
Piccirello: Remove this comment?

pkg/tui/gui/cmp_status.go
```diff
@@ -1,3 +1,18 @@
+/*
+Copyright © 2022 Doppler <support@doppler.com>
```
Piccirello: These new copyright notices should use 2023

pkg/cmd/tui.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var tuiCmd = &cobra.Command{
+	Use:   "tui",
+	Short: "Launch TUI",
+	Args:  cobra.NoArgs,
+	Run:   tui,
+}
```
Piccirello: Is there some way we can make it apparent that this is in alpha and this could disappear at any time?

Piccirello (dismissed):
I have only reviewed the code outside of the `pkg/tui` directory. Very excited to get this out into the wild.

pkg/cmd/tui.go
apazzolini: I changed the text to `Launch TUI (alpha - may change or disappear in the future)`

pkg/tui/gui/cmp_secret_view.go
```diff
@@ -0,0 +1,152 @@
...
+		OriginalName:  svm.originalName,
+		OriginalValue: svm.originalValue,
+		Name:          svm.nameView.TextArea.GetContent(),
+		Value:         svm.valueView.TextArea.GetContent(),
+		ShouldDelete:  svm.shouldDelete,
+	}
+}
+
+func (svm *SecretViewModel) ShouldSubmit() bool {
+	return svm.isDirty && (len(svm.nameView.TextArea.GetContent()) > 0 || len(svm.originalName.(string)) > 0)
```
nmanoogian: Not sure where the bug is exactly but if you have an empty secret (no name or value), it panics when you try to save

nmanoogian (commented):
Still reviewing but wanted to drop this comment

pkg/tui/gui/dispatch.go
```diff
@@ -0,0 +1,210 @@
...
+	secrets := createSecrets(computedSecrets)
+	state.SetSecrets(secrets, curProj, curConf)
+
+	gui.setIsFetching(false)
+	gui.focusComponent(gui.cmps.secrets)
+}
+
+func (gui *Gui) saveSecrets(changeRequests []models.ChangeRequest) {
+	defer recoverScreenOnCrash()
+	gui.setIsFetching(true)
```
nmanoogian, line 207: When I save, it doesn't show the "Fetching..." + spinner like it does when I press enter on a project or config. Why is that?

pkg/cmd/tui.go
Piccirello: Was thinking about this a bit more. Is there a phase in which we'll transition this from alpha to beta before GA? If not, I think we can label this command `Launch TUI (BETA)`.

pkg/tui/gui/cmp_secret_view.go
apazzolini: Good find. Fixed.

pkg/tui/gui/gui.go
```diff
@@ -0,0 +1,153 @@
+package gui
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/DopplerHQ/cli/pkg/tui/common"
+	"github.com/DopplerHQ/gocui"
+	"github.com/sasha-s/go-deadlock"
```
nmanoogian, line 25: curious: Why are we using this mutex instead of a standard RWMutex?

pkg/tui/gui/gui.go
```diff
@@ -0,0 +1,153 @@
...
+	if r := recover(); r != nil {
+		gocui.Screen.Fini()
+		panic(r)
+	}
+}
+
+func (gui *Gui) RunAndHandleError() error {
+	return SafeWithError(func() error {
+		if err := gui.run(); err != nil {
+			fmt.Println(err.Error())
```
nmanoogian: recommendation: What do you think about moving this print into the default branch so it doesn't print "quit" on normal exits?

```diff
diff --git a/pkg/tui/gui/gui.go b/pkg/tui/gui/gui.go
index 653e50e..0c58387 100644
--- a/pkg/tui/gui/gui.go
+++ b/pkg/tui/gui/gui.go
@@ -159,11 +159,11 @@ func recoverScreenOnCrash() {
 func (gui *Gui) RunAndHandleError() error {
 	return SafeWithError(func() error {
 		if err := gui.run(); err != nil {
-			fmt.Println(err.Error())
 			switch err {
 			case gocui.ErrQuit:
 				return nil
 			default:
+				fmt.Println(err.Error())
 				return err
 			}
 		}


```

nmanoogian (dismissed):
So excited to see this go out!

pkg/tui/gui/dispatch.go
apazzolini: Good catch. The save function wasn't invoked in a goroutine.

pkg/tui/gui/gui.go
apazzolini: I believe it's because this one does some amount of deadlock detection. Lazygit uses this one, so I followed suit.

apazzolini:
All feedback is addressed, re-requesting reviews for final approval. @Piccirello will defer to you for timing re:merging.

Piccirello (dismissed):
Once approved by Nic, merge at will

nmanoogian (dismissed):

apazzolini:
@Piccirello re-requesting one more time; had to rebase and resolve some Semgrep conflicts. I also went ahead and added arrow key bindings for navigating around the secret view.

Piccirello (dismissed):

pkg/tui/gui/cmp_configs.go
```diff
@@ -56,9 +57,12 @@ func CreateConfigsComponent(gui *Gui) (*ConfigsComponent, error) {
...
 	maxIdx := len(state.Configs()) - 1
-	var err error
-	self.selectedIdx, err = SelectIdx(self, idx, maxIdx)
-	return err
+	newIdx, err := SelectIdx(self, idx, maxIdx)
+	if err == nil {
+		return nil
+	}
+	self.selectedIdx = newIdx
+	return nil
```
nmanoogian, line 65: Is this logic correct? If `err` is non-nil, it doesn't look like we ever return it

pkg/tui/gui/cmp_projects.go
```diff
@@ -56,9 +57,12 @@ func CreateProjectsComponent(gui *Gui) (*ProjectsComponent, error) {
...
 	maxIdx := len(state.Projects()) - 1
-	var err error
-	self.selectedIdx, err = SelectIdx(self, idx, maxIdx)
-	return err
+	newIdx, err := SelectIdx(self, idx, maxIdx)
+	if err == nil {
+		return nil
+	}
+	self.selectedIdx = newIdx
+	return nil
```
nmanoogian, line 65: Similar logic question

nmanoogian (dismissed):
Updates look good! Can you rename all commits with `chore:` that you don't want to appear in the changelog?

pkg/tui/gui/cmp_configs.go
apazzolini: Good catch. Fixed.

pkg/tui/gui/cmp_projects.go
apazzolini: Good catch. Fixed.

apazzolini:
> Updates look good! Can you rename all commits with `chore:` that you don't want to appear in the changelog?

Done

Piccirello:
You'll probably want to prepend `chore:` to every commit other than "Add Doppler TUI" since they'll get added to the changelog. And maybe reword that to "Add tui command for editing secrets" or something like that.

apazzolini:
> You'll probably want to prepend `chore:` to every commit other than "Add Doppler TUI" since they'll get added to the changelog. And maybe reword that to "Add tui command for editing secrets" or something like that.

Done

Piccirello (approved):

nmanoogian (approved):

PR #416, opened 2023-07-06
Support updating CLI installed via winget

Piccirello:
Tested auto updates and `doppler update` command on macOS, Windows, and Git bash (MINGW64). Also verified CLI throws appropriate errors when installed via Scoop.

Closes ENG-6426

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -81,7 +81,7 @@ func CheckUpdate(command string) (bool, models.VersionCheck) {
 		return false, models.VersionCheck{}
 	}
 
-	if utils.IsWindows() {
+	if utils.IsWindows() && !utils.IsMINGW64() {
 		utils.Log(fmt.Sprintf("Update: Doppler CLI %s is available\n\nYou can update via 'scoop update doppler'\n", versionCheck.LatestVersion))
 		configuration.SetVersionCheck(versionCheck)
 		return false, models.VersionCheck{}
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -111,7 +111,7 @@ func RunInstallScript() (bool, string, Error) {
 	if !apiErr.IsNil() {
 		return false, "", Error{Err: apiErr.Unwrap(), Message: apiErr.Message}
 	}
-	fetchScriptDuration := time.Now().Sub(startTime).Milliseconds()
+	fetchScriptDuration := time.Since(startTime).Milliseconds()
 
 	CaptureEvent("InstallScriptDownloaded", map[string]interface{}{"durationMs": fetchScriptDuration})
 
@@ -139,7 +139,7 @@ func RunInstallScript() (bool, string, Error) {
 		out, err = exec.Command(command[0], command[1:]...).CombinedOutput() // #nosec G204
 	}
 
-	executeDuration := time.Now().Sub(startTime).Milliseconds()
+	executeDuration := time.Since(startTime).Milliseconds()
 
 	strOut := string(out)
 	// log output before checking error
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -24,10 +24,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DopplerHQ/cli/pkg/configuration"
 	"github.com/DopplerHQ/cli/pkg/http"
 	"github.com/DopplerHQ/cli/pkg/models"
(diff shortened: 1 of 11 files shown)


Piccirello:
Now that CLI v3.64.0 is live, we have two versions available via winget, so I was update to test the update functionality using a local build. It worked swimmingly! The only question is whether updates will work once the CLI process calling `winget upgrade` is the same one that's being replaced by the new version. This is the issue that resulted in us having to roll back update support for scoop (#119). The only way to know for sure is to ship this and test in prod.

pkg/controllers/update.go
```diff
@@ -117,8 +117,13 @@ func RunInstallScript() (bool, string, Error) {
 
 	// write script to temp file
 	tmpFile, err := utils.WriteTempFile("install.sh", script, 0555)
-	// clean up temp file once we're done with it
-	defer os.Remove(tmpFile)
+	if tmpFile != "" {
+		// clean up temp file once we're done with it
+		defer os.Remove(tmpFile)
+	}
+	if err != nil {
```
nmanoogian, line 145: check: Should this be done before the `defer`?

nmanoogian (dismissed):

pkg/controllers/update.go
Piccirello: There's no guarantee that if an error was thrown the file wasn't created. I actually see a couple edge cases in `WriteTempFile` so will put up another PR to fix those. I think this current code is safe given that the deferred removal does an empty string check.

nmanoogian (dismissed):

pkg/controllers/update.go
Piccirello: I had to rebase so pushed up an additional commit with the `WriteTempFile` improvements.

Piccirello:
@nmanoogian I pushed up some [changes](https://github.com/DopplerHQ/cli/compare/3f76a51b1dc174ed4fad70db2dbab702920a707f..09e9af3277b5864a450e5d12d26770c32f705b85) to fix a bug in the `doppler update` command. I've also added e2e tests to ensure this functionality doesn't break again.

pkg/controllers/update.go
```diff
@@ -149,3 +243,127 @@ func CLIChangeLog() (map[string]models.ChangeLog, http.Error) {
...
+	var out io.Writer = nil
+	if utils.CanLogDebug() {
+		out = os.Stderr
+	}
+	cmd, err := utils.RunCommandString(command, os.Environ(), nil, out, out, true)
+	if err != nil {
+		utils.LogDebugError(err)
+	}
+	_, err = utils.WaitCommand(cmd)
+	if err != nil {
```
nmanoogian, line 299: check: If `RunCommandString` fails but `WaitCommand` succeeds, won't it overwrite `err` with `nil`? If so, we'd correctly print the error but we'd return the incorrect bool from this function.

Other calls to `RunCommandString` appear to return out of their error handler

pkg/controllers/update.go
```diff
@@ -149,3 +243,127 @@ func CLIChangeLog() (map[string]models.ChangeLog, http.Error) {
...
+
+	var out io.Writer = nil
+	if utils.CanLogDebug() {
+		out = os.Stderr
+	}
+	cmd, err := utils.RunCommandString(command, os.Environ(), nil, out, out, true)
+	if err != nil {
+		utils.LogDebugError(err)
+	}
+	_, err = utils.WaitCommand(cmd)
```
nmanoogian, line 298: check: Do we not care about the exit code in this instance? (+`IsUpdateAvailableViaWinget`)

nmanoogian (dismissed):
Just two checks, otherwise approved!

pkg/controllers/update.go
Piccirello: Whoops, this is missing a `return false`. Good catch.

The return value technically isn't accurate since an error here doesn't necessarily mean that the CLI isn't installed via winget, but I don't have a better pattern for now. I think it's also unlikely that we'll hit this.

pkg/controllers/update.go
Piccirello: `WaitCommand()` is a wrapper around `cmd.Wait()`, which states `The returned error is nil if the command runs, has no problems copying stdin, stdout, and stderr, and exits with a zero exit status.`. So if `err` is `nil`, we know the exit code is `0`. This it shouldn't really matter.

nmanoogian (approved):

PR #418, opened 2023-07-10
Fix navigating within secret values

apazzolini:
I noticed this while using the TUI today. We last-minute added the ability to navigate secrets with arrow keys, but that causes a collision with navigating *within* secret values when editing multi-line secrets. Let's remove arrow key navigation for now as a quick-fix.

I didn't use the `chore:` prefix in the commit message because I wasn't sure if we could release without anything in the changelog, but let me know if I should add it in.


Piccirello (approved):
You can indeed use `chore:` on the commit here. This will go out with some other changes, so the changelog won't be empty.

PR #419, opened 2023-07-11
Switch to Bearer auth

Piccirello:
Our docs recommend Bearer auth due to its simplicity over Basic auth.


nmanoogian (approved):
Nice cleanup 👍

PR #420, opened 2023-07-17
Fix installing updates via winget

Piccirello:
`winget upgrade` errors if the CLI is running when we try to update it. We need to exec `winget upgrade` in the background and then immediately exit. This has the downside that our error handling/messaging isn't as robust, we can't capture timing/success anlytics, and upgrading via the auto-upgrade prompt does not result in running the original command. For ex, if a user runs `doppler configs` and then chooses to update when prompted, the CLI will update and exit before executing the `configs` command. We want users to always be on the latest CLI version, so these limitations are preferable to requiring users to manually run `doppler update`.

Closes ENG-6662

### DIFF
--- a/pkg/controllers/update.go
+++ b/pkg/controllers/update.go
@@ -86,13 +86,13 @@ func CheckUpdate(command string) (bool, models.VersionCheck) {
 	}
 
 	if utils.IsWindows() && !utils.IsMINGW64() {
-		if !InstalledViaWinget() {
+		if !installedViaWinget() {
 			utils.Log(fmt.Sprintf("Update: Doppler CLI %s is available\n\nYou can update via 'scoop update doppler'\nWe recommend installing the Doppler CLI via winget for easier updates", versionCheck.LatestVersion))
 			configuration.SetVersionCheck(versionCheck)
 			return false, models.VersionCheck{}
 		}
 
-		if !IsUpdateAvailableViaWinget(versionCheck.LatestVersion) {
+		if !isUpdateAvailableViaWinget(versionCheck.LatestVersion) {
 			CaptureEvent("UpgradeNotAvailableViaWinget", map[string]interface{}{"version": versionCheck.LatestVersion})
 			utils.LogDebug(fmt.Sprintf("Doppler CLI version %s is not yet available via winget", versionCheck.LatestVersion))
 			// reuse old version so we prompt the user again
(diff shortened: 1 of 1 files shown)


pkg/controllers/update.go
```diff
@@ -330,41 +335,17 @@ func isUpdateAvailableViaWinget(updateVersion string) bool {
...
-		} else {
-			e = Error{Err: err, Message: "Unable to upgrade via winget"}
-		}
+	c := []string{"cmd.exe", "/C", "winget", "upgrade", "--id", wingetPackageId, "--exact", "--disable-interactivity", "--version", strings.TrimPrefix(version, "v")}
 
-		CaptureEvent("WingetUpgradeFailed", map[string]interface{}{"durationMs": executeDuration})
-		return false, "", e
+	utils.LogDebug(fmt.Sprintf("Executing \"%s\"", strings.Join(c[2:], " ")))
+	cmd := exec.Command(c[0], c[1:]...)
+	if err := cmd.Run(); err != nil {
```
nmanoogian: check: Do we need to put a delay before the backgrounded `winget upgrade` command in case it starts faster than we can exit?

pkg/controllers/update.go
```diff
@@ -252,7 +252,12 @@ func InstallUpdate(version string) {
 	var controllerErr Error
 	if utils.IsWindows() && !utils.IsMINGW64() {
 		if installedViaWinget() {
-			wasUpdated, installedVersion, controllerErr = updateViaWinget(version)
+			if err := updateViaWinget(version); err != nil {
+				utils.HandleError(err, "Unable to execute winget")
+			}
+
+			utils.LogDebug("Executing winget in background, CLI is now exiting")
+			os.Exit(0)
```
nmanoogian, line 260: check: Now that we're exiting immediately, we're no longer calling `SetVersionCheck` at the bottom of this function. Does this create any problems?

pkg/controllers/update.go
```diff
@@ -329,7 +329,7 @@ func isUpdateAvailableViaWinget(updateVersion string) bool {
 	utils.LogDebug(strOut)
 
 	// Ex: `Doppler.doppler 3.63.1 3.64.0 winget`
-	re := regexp.MustCompile(fmt.Sprintf(`%s\s+%s\s+%s\s+winget`, wingetPackageId, strings.TrimPrefix(version.ProgramVersion, "v"), strings.TrimPrefix(updateVersion, "v")))
+	re := regexp.MustCompile(fmt.Sprintf(`%s\s+\d+\.\d+\.\d+\s+%s\s+winget`, wingetPackageId, strings.TrimPrefix(updateVersion, "v")))
```
nmanoogian, line 332: check: Why do we need to do this? The strict check seems more desirable

nmanoogian (dismissed):
Nice find 👍

pkg/controllers/update.go
Piccirello: Hah, spoken like someone that clearly hasn't used winget.

pkg/controllers/update.go
Piccirello: Good q. On the next CLI run we'll check for an update, see that there isn't one, and then call `SetVersionCheck`.

pkg/controllers/update.go
Piccirello: This fixes an edge case where the version of the CLI binary performing the update is different than the binary version previously installed by winget. It's mostly just to aid with testing, though I didn't see a downside from the change. If a user does hit this, we probably still want to allow them to update.

nmanoogian (approved):
Switching to `RunCommandString` looks good to me

PR #427, opened 2023-09-05
Fix forwarding of env vars when using --preserve-env

Piccirello:
Previously, we incorrectly ignored all non-preserved environment variables. We should only ignore non-preserved environment variables that are also Doppler secrets. Tests have been added for this scenario.

Closes ENG-6948

### DIFF
--- a/tests/e2e/run.sh
+++ b/tests/e2e/run.sh
@@ -140,4 +140,16 @@ beforeEach
 value="$(TEST="foo" "$DOPPLER_BINARY" run --preserve-env="INVALID" -- printenv TEST)"
 [[ "$value" == "abc" ]] || error "ERROR: existing env var not ignored when preserve-env flag passed list of nonexistent secret names"
 
+beforeEach
+
+# verify preserve-env flag preserves env vars that aren't Doppler secrets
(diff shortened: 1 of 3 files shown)


tests/e2e/run.sh
```diff
@@ -140,4 +140,16 @@ beforeEach
 value="$(TEST="foo" "$DOPPLER_BINARY" run --preserve-env="INVALID" -- printenv TEST)"
 [[ "$value" == "abc" ]] || error "ERROR: existing env var not ignored when preserve-env flag passed list of nonexistent secret names"
 
+beforeEach
+
+# verify preserve-env flag preserves env vars that aren't Doppler secrets
+value="$(NOT_DOPPLER_SECRET="foo" "$DOPPLER_BINARY" run --preserve-env="TEST" -- printenv NOT_DOPPLER_SECRET || true)"
+[[ "$value" == "foo" ]] || error "ERROR: existing env var not preserved when preserve-env flag passed unrelated secret name"
```
nmanoogian, line 147: check: The logic looks good but are both of these error messages correct?

nmanoogian (dismissed):

tests/e2e/run.sh
Piccirello: Copy pasta

nmanoogian (approved):

PR #428, opened 2023-09-07
add workflow for vulncheck

seslattery:
From https://go.dev/security/vuln/ : 
Integrations with [pkg.go.dev](https://pkg.go.dev/) and govulncheck to enable developers to find vulnerabilities in their projects. The [govulncheck command](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) analyzes your codebase and only surfaces vulnerabilities that actually affect you, based on which functions in your code are transitively calling vulnerable functions. Govulncheck provides a low-noise, reliable way to find known vulnerabilities in your projects.

### DIFF
--- a/.github/workflows/vulncheck.yml
+++ b/.github/workflows/vulncheck.yml
@@ -1,0 +1,25 @@
+name: Vulncheck
+
+on: [pull_request, push]
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  vulncheck:
+    name: Analysis
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+      - name: Set up Go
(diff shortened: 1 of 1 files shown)


Piccirello (approved):
Very nice!

nmanoogian (approved):

.github/workflows/vulncheck.yml
```diff
@@ -0,0 +1,25 @@
...
+  vulncheck:
+    name: Analysis
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.21.0
```
Piccirello, line 18: Just realized you'll want to omit the patch version here - so just `1.21`

PR #433, opened 2023-10-06
chore: provide better error message for TLS error

Piccirello:
Closes ENG-7188


watsonian:
@Piccirello Would there be any value to calling out specifically that TLS v1.3 is required?

Piccirello:
> @Piccirello Would there be any value to calling out specifically that TLS v1.3 is required?

It will be once we make that change, but it isn't yet. We currently requires TLS v1.2 and AFAIK no one has hit this error under that requirement.

watsonian (approved):

PR #436, opened 2023-10-10
Add `configure flags` command for configuring CLI behavior

### LINKED ISSUE #407: [FEATURE] Allow user to disable or silence `valueFromEnvironmentNotice` warnings
**Is your feature request related to a problem? Please describe.**
When using environment variables to configure Doppler, the `valueFromEnvironmentNotice` CLI function writes a warning to stderr each time a Doppler command is executed. This behavior can be irritating, especially when executing `doppler run` dozens of times per day. 

I understand that these messages were originally printed to stdout, but due to the "noisiness" of the output, they were later changed to stderr. I think that solved one issue while creating the same issue somewhere else. These messages are warnings, not output or errors.  

**Describe the solution you'd like**
There are a couple of different approaches that would work:
1. It would be helpful if users could disable or silence the `valueFromEnvironmentNotice` warnings, either by executing a config command and storing the preference, setting an environment variable, or warning once per session (or day) and requiring the user to agree on the CLI with `y/n` before continuing.
2. Reduce the frequency of the warning output to once per session, or once per day
3. Make these warnings opt-in. I'm not sure these warnings should really be on by default. I think I understand the use case – to alert the user that a "hidden" flag is overriding the CLI – but isn't almost everything about Doppler's config "hidden" from the user on the CLI? These special environment variables are a supported and documented way to configure Doppler, and any user who happens to have set them, should know that they are set. So perhaps they should be opt-in instead with a flag `--verbose`, `--debug`, or `--enable-warnings`?

**Describe alternatives you've considered**
I've tried wrapping the doppler CLI binary in a shell function that redirects all stderr output to /dev/null. This quickly became overly complicated and required additional logic to ensure "valid" stderr actually was emitted.

**Additional context**
This behavior creates other unintended consequences. This issue was raised in connection with a related issue with the Doppler VSCode extension (https://github.com/DopplerHQ/vscode/issues/22). Allowing users to disable or silence these warning provides greater flexibility and usability for customers who prefer to use environment variables to configure Doppler.

### LINKED ISSUE #422: [FEATURE] Configure doppler repo-wide to stop asking to update itself
**Is your feature request related to a problem? Please describe.**
In my project we use [`asdf`](https://github.com/takutakahashi/asdf-doppler) to install specific versions of the tools we use, so that all of the members of our team are using the same version tools.

But `doppler` prompts individuals to install the latest version by default. There's a flag to prevent this, but then we have to make sure as a team to always provide that flag whenever we invoke the CLI, which is impractical beyond a couple people.

**Describe the solution you'd like**
A way to configure centrally in a repository that we don't want to check for updates, since we're managing the version separately.

**Describe alternatives you've considered**
Adding `--no-check-version` or the equivalent environment variable to every invocation of Doppler in my monorepo, but that's a game of whack-a-mole trying to make sure that in a large sprawling monorepo that we don't ever miss it.

Piccirello:
This PR adds the ability to configure various CLI behavior via a new `configure flags` command. The following flags default to `true` and can be disabled via `configure flags disable` or via `setup` with a valid `doppler.yaml`.

- `analytics`: Whether anonymous analytics are enabled.
- `env-warning`: Whether to print a warning when CLI behavior is modified via an environment variable.
- `update-check`: Whether to check for CLI updates.

Closes ENG-6729, closes #422, closes #407


pkg/cmd/root.go
```diff
@@ -61,6 +62,21 @@ var rootCmd = &cobra.Command{
...
+		if version.PerformVersionCheck && configuration.CanReadEnv {
+			enable := os.Getenv("DOPPLER_ENABLE_VERSION_CHECK")
+			if enable == "false" {
+				logValueFromEnvironmentNotice("DOPPLER_ENABLE_VERSION_CHECK")
+				version.PerformVersionCheck = false
+			}
+		}
+		if version.PerformVersionCheck {
+			version.PerformVersionCheck = !utils.GetBoolFlagIfChanged(cmd, "no-check-version", !version.PerformVersionCheck)
+		}
```
Piccirello, line 78: Relocated this logic from `loadFlags()` since the config file isn't loaded yet when that function runs.

pkg/cmd/configure_flags.go
```diff
@@ -0,0 +1,151 @@
...
+func FlagsValidArgs(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	persistentValidArgsFunction(cmd)
+
+	return models.GetFlags(), cobra.ShellCompDirectiveNoFileComp
+}
+
+func init() {
+	configureCmd.AddCommand(configureFlagsCmd)
+
+	configureFlagsGetCmd.Flags().Bool("plain", false, "print values without formatting. values will be printed in the same order as specified")
```
nmanoogian: Two small UX things here. This says `values will be printed in the same order...` but the command only takes one flag. The usage text shows:

```
➜  cli git:(tom-flags) ./doppler flags get env-warning analytics
Error: accepts 1 arg(s), received 2
Usage:
  doppler flags get [flag] [flags]
```

In that usage message, `[flag]` is referring to the Doppler flag name (e.g. `analytics`) but `[flags]` is referring to CLI flags used to augment the command. This is a bit confusing but I think it's a nofix; most people will get it

nmanoogian (dismissed):
Loving this impl, glad you went this direction 👍 

I did some light testing locally but your E2Es look pretty comprehensive!

pkg/cmd/configure_flags.go
Piccirello: Good catch on the help text, that was copy/pasta.

The usage text is indeed a bit confusing. I'm not sure how we could make it clearer other than choosing a different name for our "flags". Agreed this is a no-fix.

nmanoogian (dismissed):

agilesteel:
Any updates here? The issues this PR closes are ~8 months old, so it would be really great to merge this one in.

nmanoogian:
> Any updates here? The issues this PR closes are ~8 months old, so it would be really great to merge this one in.

Thanks for the bump, @agilesteel! I'll be taking this PR over and I should have it fixed up and ready to go shortly.

nmanoogian (approved):

apazzolini (approved):

agilesteel:
Bro for real we've been waiting for 8 months and you fixed it in 2 days. Epic! 👏🏼

nmanoogian:
> Bro for real we've been waiting for 8 months and you fixed it in 2 days. Epic! 👏🏼

Full credit goes to @Piccirello! He parted ways with Doppler before he could get this merged and all that was missing was fixing a small but tricky issue in the tests.

The delay since late October is on me; I apologize for that and we have improved our processes to make sure that PRs aren't allowed to sit for this long in the future.

Piccirello:
@nmanoogian thank you for getting this over the finish line. I see this capability potentially serving future customization needs, beyond just the several people already asking for it.

PR #443, opened 2023-11-28
Switch from alpine to distroless, add -extldflags static

seslattery:


apazzolini (commented):
Have you been able to test this locally?

PR #445, opened 2023-12-22
Update golang.org/x/crypto to v0.17.0 and tidy

nmanoogian:
Closes ENG-7453

PR #448, opened 2024-02-09
Patch Fix Insufficient Verification of Data Authenticity Improper Validation of Integrity Check Value

lamrecognitions:
Terrapin is a prefix truncation attack targeting the SSH protocol. More precisely, Terrapin breaks the integrity of SSH's secure channel. By carefully adjusting the sequence numbers during the handshake, an attacker can remove an arbitrary amount of messages sent by the client or server at the beginning of the secure channel without the client or server noticing it.

CVE-2023-48795
CWE-345
CWE-354
`CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:H/A:N`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/spf13/cobra v1.7.0
 	github.com/stretchr/testify v1.8.1
 	github.com/zalando/go-keyring v0.2.1
-	golang.org/x/crypto v0.1.0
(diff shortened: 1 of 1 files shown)


nmanoogian:
Thanks for filing this, @lamrecognitions! However, this change is already included in #445, which is currently in review.

PR #450, opened 2024-02-22
Deprecate "secrets notes set" config flag

rgharris:
Release note: Notes have always been set at the project level so it is more accurate to not provide a config. Use the project note API endpoint if config is not provided. You must still have write access to a config with the secret in order to set its note.

This depends on unreleased API changes so this will not run against our production environment today.

Closes ENG-7581


nmanoogian (approved):
Looks good! Did some light testing. Let's hold the merge here until the endpoint is fully released.

PR #457, opened 2024-05-16
Add support for value types

apazzolini:
This adds read/write support for value types in the CLI.

Notably, this destructures the `type` field inside of the value type properties for direct usage. If we ever support additional metadata on value types (e.g. regex with a pattern), we'll add a second CLI flag for setting it. This keeps the 99% case cleaner.

Closes ENG-7909

### DIFF
--- a/pkg/cmd/enclave_secrets.go
+++ b/pkg/cmd/enclave_secrets.go
@@ -105,6 +105,7 @@ func init() {
 	}
 	enclaveSecretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
 	enclaveSecretsCmd.Flags().Bool("visibility", false, "include secret visibility in table output")
+	enclaveSecretsCmd.Flags().Bool("type", false, "include secret type in table output")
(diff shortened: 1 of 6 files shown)


pkg/cmd/enclave_secrets.go
```diff
@@ -105,6 +105,7 @@ func init() {
 	}
 	enclaveSecretsCmd.Flags().Bool("raw", false, "print the raw secret value without processing variables")
 	enclaveSecretsCmd.Flags().Bool("visibility", false, "include secret visibility in table output")
+	enclaveSecretsCmd.Flags().Bool("value-type", false, "include secret value type in table output")
```
Piccirello: Thoughts on exposing this publicly as `type`? I think it's implicit that it refers to the value, given that the "secret type" is referred to as "visibility".

apazzolini: Hey Tom! I can get behind that. 👍

kyle-mcguire (dismissed):

nmanoogian (approved):
Nice! I did some light local testing and it feels great

I agree with the decision to keep any future type parameters (e.g. regex) in a separate flag

PR #458, opened 2024-06-05
Add support for CLI-based MFA recovery

nmanoogian:
This PR adds a CLI command which simply calls `POST /v3/me/mfa_recovery`.

This depends on https://github.com/DopplerHQ/server/pull/6727

### DIFF
--- a/.github/workflows/release-tests.yaml
+++ b/.github/workflows/release-tests.yaml
@@ -27,7 +27,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v1
       with:
         version: latest
-        args: release --snapshot --skip-publish --skip-sign
+        args: release --snapshot --skip=publish,sign
       env:
         GOPATH: ${{ runner.workspace }}
     - uses: actions/upload-artifact@master
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,3 +1,4 @@
+version: 2
 project_name: doppler
 
 before:
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ test-packages:
 	./tests/packages.sh
 
 test-release:
-	doppler run -- goreleaser release --snapshot --skip-publish --clean --parallelism=4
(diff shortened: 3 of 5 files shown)


pkg/cmd/mfa-recovery.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/DopplerHQ/cli/pkg/configuration"
+	"github.com/DopplerHQ/cli/pkg/http"
+	"github.com/DopplerHQ/cli/pkg/utils"
+	"github.com/spf13/cobra"
+)
+
+var mfaRecoveryCmd = &cobra.Command{
+	Use:   "mfa-recovery",
```
apazzolini: Should we make the command `doppler mfa recovery` for more symmetry with the other options? (And in case we want to do other things with mfa in the future, e.g. allow the CLI to itself be an mfa option).

apazzolini (dismissed):
Will defer to your judgment on the comment.

pkg/cmd/mfa-recovery.go
Piccirello: You could also potentially nest it under `doppler login` (i.e. `doppler login mfa-recovery`).

nmanoogian: I like both of these options! I'll go with `doppler mfa recovery` as I like the idea of having more MFA commands in the future

nmanoogian:
We're seeing an issue with our Windows install.sh test in that the response code of the download request is coming back "000" with an error code from curl. It's pretty easy to test with:

```
curl --tlsv1.2 --proto "=https" -w "%{http_code}" --silent --retry 5 -o "binary" -LN -D - "https://cli.doppler.com/download?os=windows&arch=amd64&format=zip" 2>&1
```

Thanks to @watsonian for finding this issue: https://github.com/curl/curl/issues/13845 which appears to be the same problem.

I've removed the Windows test as a required check for now and I'll monitor the curl issue for updates.

apazzolini (approved):

PR #463, opened 2024-08-29
Resolve the error - Doppler Error: exec: "x": cannot run executable found relative to current directory'

### LINKED ISSUE #415: [BUG] Cannot run executable found relative to current directory
**Describe the bug**

`doppler run COMMAND` fails if `COMMAND` is resolved from PATH and located relative to the current directory.

**To Reproduce**
Steps to reproduce the behavior (minimal example):

1. Create a `hello-world` bash script in `./bin` and add the folder to PATH:

```shell
$ mkdir bin
$ cat <<EOF > ./bin/hello-world
#!/usr/bin/env bash
echo "Hello world"
EOF

$ chmod +x ./bin/hello-world
$ export PATH=./bin:$PATH
```

2. Verify that we can run it directly:

```shell
$ hello-world
Hello world
```

3. Try running it through Doppler:

```shell
$ doppler run hello-world
Doppler Error: exec: "hello-world": cannot run executable found relative to current directory
```

Note that it does work to run `doppler run ./bin/hello-world`, not just `doppler run hello-world`.

**Expected behavior**

It should work to run commands from PATH relative to the current directory.

**CLI Version:**
Version v3.60.1

**Additional context**

Related bug with fix? https://github.com/syncthing/syncthing/issues/8499

TeddiO:
Hi there,

Starting with Go 1.19 the Go team put in some guardrails surrounding `exec.Command` and `exec.LookPath` where including the current path was considered undesirable behaviour (https://pkg.go.dev/os/exec@master#hdr-Executables_in_the_current_directory) . This PR aims to restore pre 1.19 behaviour.

The reasoning behind this is there are scripts and systems that do exist (and are created even today) that rely on this behaviour. For something such as Doppler which is essentially injecting the secrets and calling a shell (that doesn't restrict this behaviour), these guardrails seem somewhat irrelevant. 

This would resolve https://github.com/DopplerHQ/cli/issues/415 .

### DIFF
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -108,6 +108,11 @@ func Cwd() string {
 // RunCommand runs the specified command
 func RunCommand(command []string, env []string, inFile io.Reader, outFile io.Writer, errFile io.Writer, forwardSignals bool) (*exec.Cmd, error) {
 	cmd := exec.Command(command[0], command[1:]...) // #nosec G204 nosemgrep: semgrep_configs.prohibit-exec-command
+	// Resolves https://github.com/DopplerHQ/cli/issues/415
+	if errors.Is(cmd.Err, exec.ErrDot) {
+		cmd.Err = nil
+	}
+
 	cmd.Env = env
 	cmd.Stdin = inFile
 	cmd.Stdout = outFile


nmanoogian (approved):
Thanks for tracking this down, @TeddiO!

PR #480, opened 2025-02-18
Add workflow to update Salus GoVersionScanner min_version

watsonian:
From https://go.dev/doc/devel/release:

> Each major Go release is supported until there are two newer major releases.
> For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was
> supported until the Go 1.8 release. We fix critical problems, including critical
> security problems, in supported releases as needed by issuing minor revisions
> (for example, Go 1.6.1, Go 1.6.2, and so on).

We want to ensure the Salus scan keeps the GoVersionScanner's `min_version` at the latest version of the prior release series (i.e., if the most recent Go release is 1.26.3, we want it set to the most recent 1.25.x version). This workflow runs a script that fetches the latest stable releases that are currently being maintained, determines what that latest prior release version is and compares it to what we have in our Salus config. If they don't match, it updates the file and opens a PR with the change.


watsonian:
Going to take another approach.

PR #487, opened 2025-04-29
Fixed issue preventing fallback and metadata files from being saved in DOPPLER_CONFIG_DIR

nmanoogian:
The `DOPPLER_CONFIG_DIR` was being appropriately used for reading/writing the configuration files but was not being used as the default for fallback and metadata files. This change hoists the default metadata and fallback directories to the `configuration` controller so it can be defined in the same spot as `configuration.UserConfigFile`.

### DIFF
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -37,6 +37,12 @@ var UserConfigDir string
 // UserConfigFile (e.g. /home/user/doppler/.doppler.yaml)
 var UserConfigFile string
 
+// UserFallbackDir (e.g. /home/user/doppler/.doppler.yaml)
+var UserFallbackDir string
+
+// UserMetadataDir the directory containing metadata files
+var UserMetadataDir string
+
 // Scope to use for config file
 var Scope = "."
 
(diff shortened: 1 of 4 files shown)


pkg/configuration/config.go
```diff
@@ -37,6 +37,12 @@ var UserConfigDir string
 // UserConfigFile (e.g. /home/user/doppler/.doppler.yaml)
 var UserConfigFile string
 
+// UserFallbackDir (e.g. /home/user/doppler/.doppler.yaml)
+var UserFallbackDir string
```
apazzolini, line 41: Would these be better named as `FallbackDir` and `MetadataDir` since they're still used even if the user hasn't configured them? Or am I misunderstanding that part?

apazzolini (approved):

pkg/configuration/config.go
nmanoogian: Good question -- `SetConfigDir` is called in the `init` function with the default `$HOME/.doppler` directory, which is invoked automatically when the module loads. `SetConfigDir` already updates `UserConfigDir` and `UserConfigFile` so I'm mostly just following the existing naming pattern

https://github.com/DopplerHQ/cli/blob/f383368be6389f1f4236abae9d1e12c00e8f2818/pkg/configuration/config.go#L51-L58

Mikhai56 (commented):
Vanros zakrit

PR #489, opened 2025-05-27
Alias --fallback-only with --offline

### LINKED ISSUE #465: [FEATURE] Offline mode
**Is your feature request related to a problem? Please describe.**
Offline development

**Describe the solution you'd like**
Add an "offline" flag where the CLI doesn't try to fetch the current secrets (every request tries 5 times by default with 10s timeout), but just immediately uses the locally stored secrets. Other solution could be detection that the computer is offline and don't attempt to fetch them automatically.

**Describe alternatives you've considered**
Setting --timeout and --attempts to low amounts, but this is a bit cumbersome

**Additional context**
When developing on planes without wifi etc. this would be nice.

rgharris:
This flag can be useful when using the CLI offline as network requests are not made when only using the fallback file. These changes apply to `doppler run` and `doppler secrets download`.

The approach taken here seems cleaner to me than the one originally taken in #478. Plus, the error messages here are clearer as they mention the specific flag (`--fallback-only` or `--offline`) that you utilized:

```
> doppler run --project=aaa --config=stg --scope=dev --offline --no-fallback --watch echo
Warning: --offline has no effect when the fallback file is disabled
Warning: --watch has no effect when used with --offline
Doppler Error: Conflict: unable to specify --no-fallback with --offline

> doppler run --project=aaa --config=stg --scope=dev --fallback-only --no-fallback --watch echo
Warning: --fallback-only has no effect when the fallback file is disabled
Warning: --watch has no effect when used with --fallback-only
Doppler Error: Conflict: unable to specify --no-fallback with --fallback-only
```

Closes #465
Closes ENG-8748

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -78,6 +78,12 @@ doppler run --mount secrets.json -- cat secrets.json`,
 		enableCache := enableFallback && !utils.GetBoolFlag(cmd, "no-cache")
 		fallbackReadonly := utils.GetBoolFlag(cmd, "fallback-readonly")
 		fallbackOnly := utils.GetBoolFlag(cmd, "fallback-only")
+		var fallbackFlag string
+		if cmd.Flags().Changed("offline") {
+			fallbackFlag = "--offline"
+		} else {
+			fallbackFlag = "--fallback-only"
+		}
 		exitOnWriteFailure := !utils.GetBoolFlag(cmd, "no-exit-on-write-failure")
 		preserveEnv := cmd.Flag("preserve-env").Value.String()
 		forwardSignals := utils.GetBoolFlag(cmd, "forward-signals")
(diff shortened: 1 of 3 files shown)


pkg/cmd/enclave_secrets.go
```diff
@@ -167,7 +167,8 @@ func init() {
 	enclaveSecretsDownloadCmd.Flags().Bool("no-fallback", false, "disable reading and writing the fallback file")
 	enclaveSecretsDownloadCmd.Flags().String("fallback-passphrase", "", "passphrase to use for encrypting the fallback file. by default the passphrase is computed using your current configuration.")
 	enclaveSecretsDownloadCmd.Flags().Bool("fallback-readonly", false, "disable modifying the fallback file. secrets can still be read from the file.")
-	enclaveSecretsDownloadCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
+	enclaveSecretsDownloadCmdFallbackOnly := enclaveSecretsDownloadCmd.Flags().Bool("fallback-only", false, "read all secrets directly from the fallback file, without contacting Doppler. secrets will not be updated. (implies --fallback-readonly)")
```
nmanoogian: The `enclave` commands are deprecated so we can leave these unchanged

nmanoogian (dismissed):
Nice! Tested lightly and works as expected

pkg/cmd/enclave_secrets.go
rgharris: [Force push](https://github.com/DopplerHQ/cli/compare/94b7f94a728ad5db03e49b5b756539db4c071b82..157a031786abe66609876f7f02c99016905de19e) removes the changes to the `enclave secrets download` command. Tested and works as expected - `Flags().Changed()` doesn't complain that the flag wasn't defined for the command.

nmanoogian (approved):

PR #491, opened 2025-06-17
Reconnect to Doppler after network failure when using --watch

apazzolini:
Prior to this PR, if a client is using the `--watch` flag and loses connectivity to the Doppler API (either because they lost internet or the service they were connected to went down), the process started by `doppler run` would continue running but never be restarted.

This PR will now detect connectivity failures and handle reconnection. When reconnecting, the process will only be restarted if secrets have indeed changed; if no secrets were changed, the pre-existing process will continue running.

Closes ENG-8855

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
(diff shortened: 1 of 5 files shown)


apazzolini:
Recreating this PR, I think GH was having issues when it was created.

PR #492, opened 2025-06-17
 Reconnect to Doppler after network failure when using --watch

apazzolini:
Prior to this PR, if a client is using the --watch flag and loses connectivity to the Doppler API (either because they lost internet or the service they were connected to went down), the process started by doppler run would continue running but never be restarted.

This PR will now detect connectivity failures and handle reconnection. When reconnecting, the process will only be restarted if secrets have indeed changed; if no secrets were changed, the pre-existing process will continue running.

Closes ENG-8855

### DIFF
--- a/pkg/utils/number.go
+++ b/pkg/utils/number.go
@@ -15,21 +15,25 @@ limitations under the License.
 */
 package utils
 
-func Min(x, y int) int {
+import (
+	"golang.org/x/exp/constraints"
+)
+
+func Min[T constraints.Ordered](x, y T) T {
 	if x < y {
 		return x
 	}
 	return y
 }
 
-func Max(x, y int) int {
+func Max[T constraints.Ordered](x, y T) T {
 	if x > y {
 		return x
 	}
 	return y
 }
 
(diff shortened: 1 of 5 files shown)


pkg/utils/number.go
```diff
@@ -15,13 +15,22 @@ limitations under the License.
...
+import "time"
+
 func Min(x, y int) int {
 	if x < y {
 		return x
 	}
 	return y
 }
 
+func MinDuration(x, y time.Duration) time.Duration {
```
emily-curry: optional: I think go's generics might be useful here:
```suggestion
func Min[T constraints.Ordered](x, y T) T {
```

emily-curry (dismissed):
Good stuff! I tested locally, tried to break it, but this feels pretty solid. I'm going to keep using this version locally for a while to see if I get any unexpected restarts, but this is good to merge now.

apazzolini:
@emily-curry Applied the generic change, note that the CLI repo requires re-review when anything changes.

emily-curry (approved):

PR #494, opened 2025-06-21
Atomically create and check existence of configuration and fallback directories

### LINKED ISSUE #493: [BUG] Race condition when running doppler concurrently on CI "Unable to create config directory"
**Describe the bug**
When running doppler concurrently in GitHub actions, I'm occasionally seeing "Unable to create config directory" or "Unable to create directory for fallback file".

**To Reproduce**
Steps to reproduce the behavior. Please include output from running the command with `--debug`.

In bash, run:
```bash
export DP="doppler --no-check-version"
rm -rf ~/.doppler
$DP me & $DP me
```

Do that a few times and you will see an error message like one of the following:

```
Unable to create config directory /Users/doug/.doppler
Doppler Error: mkdir /Users/doug/.doppler: file exists
```

**Expected behavior**
I should be able to run concurrent doppler commands in CI (when I have a DOPPLER_TOKEN) set without error.

**Screenshots**

Image

**CLI Version:**
Version v3.72.0

**Additional context**

The problem is that the doppler CLI does an existence check and then as a separate step creates the directory in two places.

In `pkg/configuration/config.go`:

```go
	if !utils.Exists(UserConfigDir) {
		utils.LogDebug(fmt.Sprintf("Creating the config directory %s", UserConfigDir))
		err := os.Mkdir(UserConfigDir, 0700)
		if err != nil {
			utils.HandleError(err, fmt.Sprintf("Unable to create config directory %s", UserConfigDir))
		}
	}
```

In `pkg/cmd/run.go`:

```go
		if !utils.Exists(configuration.UserFallbackDir) {
			err := os.Mkdir(configuration.UserFallbackDir, 0700)
			if err != nil {
				utils.LogDebug("Unable to create directory for fallback file")
				if exitOnWriteFailure {
					utils.HandleError(err, "Unable to create directory for fallback file", strings.Join(controllers.WriteFailureMessage(), "\n"))
				}
			}
		}
```

drichardson:
There is a race condition when doppler CLI is run concurrently.

This PR creates/checks existence of files in one syscall (mkdir) to avoid this race condition for the configuration (`~/.doppler`) and fallback (`~/.doppler/fallback`) directories.

Closes #493 

Repeated test in #493 with this change and could no longer reproduce the race condition.

### DIFF
--- a/pkg/cmd/run.go
+++ b/pkg/cmd/run.go
@@ -564,13 +564,11 @@ func initFallbackDir(cmd *cobra.Command, config models.ScopedOptions, format mod
 			legacyFallbackPath = legacyFallbackFile(config.EnclaveProject.Value, config.EnclaveConfig.Value)
 		}
 
-		if !utils.Exists(configuration.UserFallbackDir) {
-			err := os.Mkdir(configuration.UserFallbackDir, 0700)
-			if err != nil {
-				utils.LogDebug("Unable to create directory for fallback file")
-				if exitOnWriteFailure {
-					utils.HandleError(err, "Unable to create directory for fallback file", strings.Join(controllers.WriteFailureMessage(), "\n"))
-				}
+		err := os.Mkdir(configuration.UserFallbackDir, 0700)
+		if err != nil && !os.IsExist(err) {
+			utils.LogDebug("Unable to create directory for fallback file")
+			if exitOnWriteFailure {
+				utils.HandleError(err, "Unable to create directory for fallback file", strings.Join(controllers.WriteFailureMessage(), "\n"))
 			}
 		}
 	}
--- a/pkg/configuration/config.go
+++ b/pkg/configuration/config.go
@@ -70,12 +70,9 @@ func Setup() {
 	utils.LogDebug(fmt.Sprintf("Using config dir %s", UserConfigDir))
 	utils.LogDebug(fmt.Sprintf("Using config file %s", UserConfigFile))
 
-	if !utils.Exists(UserConfigDir) {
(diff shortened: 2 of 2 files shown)


nmanoogian:
Thank you for investigating this and putting up a fix! Your changes look good and do indeed fix the issue in my testing.

Looking at your report, I suspect that there might be other race-condition-related bugs in concurrent runs of commands that modify the config file (`doppler setup` and `doppler login`) -- we might need to consider some kind of file-based locking to resolve those completely. However, I think the directory creation is going to be one of the largest areas of concern, especially in CI where lots of scripting/concurrency is happening. Moreover, your fix isn't likely to obscure any of those config-related race conditions, which could still happen today anyway (without your fix) if the directories are already in place.

Two quick asks:

- For better or worse, our commit messages are automatically turned into release notes during our deploys. You can include additional information in the commit body but can you revise your commit message to something like this?
  - "Fix race condition in concurrent creation of config and fallback dirs"
- It looks like we have an e2e test failing here. These can be irritating to pin down with our testing harness so I'll look into this and get back to you with a fix unless you beat me to it

drichardson:
Yeah I can fix the commit message. I'm on vacay rn and didn't bring my computer with me so will be a couple weeks. 

Or feel free to rewrite my commit message, or just throw this PR in the trash and fix it elsewhere.

nmanoogian:
Sounds good! I'll push your changes back up onto an internal branch. I'd typically prefer that we do it through an external PR so you can get attribution but I think that E2E test is also failing because of a secrets access issue related to the PR being external.

nmanoogian:
Closing in favor of https://github.com/DopplerHQ/cli/pull/495

PR #503, opened 2025-11-19
fix: handle broken pipe error while rapidly reading from secrets mount

### LINKED ISSUE #502: [BUG] `vite dev` server sometimes crashing with broken pipe error
**Describe the bug**
`vite dev` running under `doppler run` sometimes crashes from a broken pipe error.

**To Reproduce**
Unsure exactly how to get it do this, but it only happens when vite seemingly gets into an endless loop of reading the secrets mount:

Image

**Expected behavior**
`vite dev` does not crash

 - OS: Linux
 - Version Fedora 43

**CLI Version:**
Version 3.75.1

hmnd:
potentially fixes #502

I've been running this without the crash for about half an hour, so I *think* it solves it. Hard to tell, because I'm still getting spammed by the *Secrets mount opened by reader* message, but I don't think this would introduce any further breaking behavior either.

### DIFF
--- a/pkg/controllers/secrets.go
+++ b/pkg/controllers/secrets.go
@@ -225,6 +225,9 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 				utils.HandleError(err, message)
 			}
 
+			if enableReadsLimit {
+				numReads++
+			}
 			utils.LogDebug("Secrets mount opened by reader")
 
 			if _, err := f.Write(secrets); err != nil {
@@ -259,9 +262,6 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 				cleanupFIFO()
 				utils.HandleError(err, message)
 			}
-
-			// only increment read count after successfully writing and closing
-			numReads++
 
 			// delay before re-opening file so reader can detect an EOF.
 			// if we immediately re-open the file, the original reader will keep reading
--- a/pkg/controllers/secrets.go
+++ b/pkg/controllers/secrets.go
@@ -236,8 +236,7 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 					break
 				}
 				// broken pipe occurs when reader closes pipe before writing completes (eg. with vite dev server)
-				var errno syscall.Errno
-				if errors.As(err, &errno) && errno == syscall.EPIPE {
+				if errors.Is(err, syscall.EPIPE) {
 					utils.LogDebug("Reader closed pipe before write completed")
 					_ = f.Close()
 					continue
@@ -253,8 +252,7 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
 					break
 				}
 				// broken pipe on close is safe to ignore - the reader has already disconnected
-				var errno syscall.Errno
-				if errors.As(err, &errno) && errno == syscall.EPIPE {
+				if errors.Is(err, syscall.EPIPE) {
 					utils.LogDebug("Pipe closed by reader")
 					continue
 				}
--- a/pkg/controllers/secrets.go
+++ b/pkg/controllers/secrets.go
@@ -26,6 +26,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/secrets.go
```diff
@@ -241,10 +249,20 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
...
+					utils.LogDebug("Pipe closed by reader")
+					continue
+				}
+
 				cleanupFIFO()
 				utils.HandleError(err, message)
 			}
 
+			// only increment read count after successfully writing and closing
+			numReads++
```
Piccirello: This now allows partial writes, including writes up to n-1 bytes, without counting it as a write. I think it would be preferable to count any exposure of secrets (partial or otherwise) as a write.

hmnd: Good point! I've moved it back to where it was, but changed to only increment when max reads is set :)

hmnd:
I left a dev session running on this branch for the past ~24 hrs and it's still running so I think this is indeed a viable fix! 

I'm still confused what vite is doing, as it seems the endless *Secrets mount opened by reader* eventually stop... maybe a separate issue to file with vite.

pkg/controllers/secrets.go
```diff
@@ -224,14 +225,24 @@ func MountSecrets(secrets []byte, mountPath string, maxReads int) (string, func(
...
 			utils.LogDebug("Secrets mount opened by reader")
 
 			if _, err := f.Write(secrets); err != nil {
 				// race: cleanup has already begun; no need to error
 				if errors.Is(err, fs.ErrNotExist) && fifoCleanupStarted {
 					break
 				}
+				// broken pipe occurs when reader closes pipe before writing completes (eg. with vite dev server)
+				var errno syscall.Errno
+				if errors.As(err, &errno) && errno == syscall.EPIPE {
```
Piccirello: Is there a reason for not using the simpler `errors.Is`?

```suggestion
				if errors.Is(err, syscall.EPIPE) {
```

hmnd: Ahh I believe that was left over from when I was originally trying to debug this and was having trouble with errors.Is on the wrong syscall error. Fixed now.

mikesellitto (dismissed):
Looks good. I tested locally with the following test which fails on `master` and succeeds with the proposed changes (feel free to commit this if you care to).

@seslattery thoughts?

```diff --git a/pkg/controllers/secrets_test.go b/pkg/controllers/secrets_test.go
index bccb47c..bdd2e78 100644
--- a/pkg/controllers/secrets_test.go
+++ b/pkg/controllers/secrets_test.go
@@ -17,9 +17,13 @@ limitations under the License.
 package controllers
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/DopplerHQ/cli/pkg/utils"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -106,3 +110,61 @@ func TestSecretsToBytes(t *testing.T) {
 		t.Errorf("Unable to convert secrets to byte array in %s format", format)
 	}
 }
+
+func TestMountSecrets(t *testing.T) {
+	if !utils.SupportsNamedPipes {
+		t.Skip("Named pipes not supported on this platform")
+	}
+
+	secrets := []byte(`{"SECRET_KEY":"secret_value"}`)
+	mountPath := filepath.Join(t.TempDir(), "secrets_mount")
+
+	path, cleanup, err := MountSecrets(secrets, mountPath, 1)
+	if !err.IsNil() {
+		t.Fatalf("MountSecrets failed: %v", err.Err)
+	}
+	defer cleanup()
+
+	time.Sleep(50 * time.Millisecond)
+
+	content, readErr := os.ReadFile(path)
+	assert.NoError(t, readErr)
+	assert.Equal(t, string(secrets), string(content))
+}
+
+func TestMountSecretsBrokenPipe(t *testing.T) {
+	if !utils.SupportsNamedPipes {
+		t.Skip("Named pipes not supported on this platform")
+	}
+
+	// large data to exceed pipe buffer and trigger EPIPE when reader closes early
+	secrets := make([]byte, 256*1024)
+	for i := range secrets {
+		secrets[i] = byte('A' + (i % 26))
+	}
+	mountPath := filepath.Join(t.TempDir(), "secrets_mount_epipe")
+
+	path, cleanup, err := MountSecrets(secrets, mountPath, 0)
+	if !err.IsNil() {
+		t.Fatalf("MountSecrets failed: %v", err.Err)
+	}
+	defer cleanup()
+
+	time.Sleep(50 * time.Millisecond)
+
+	for i := 0; i < 10; i++ {
+		f, openErr := os.OpenFile(path, os.O_RDONLY, 0)
+		if openErr != nil {
+			continue
+		}
+		f.Read(make([]byte, 1))
+		f.Close()
+		time.Sleep(5 * time.Millisecond)
+	}
+
+	time.Sleep(100 * time.Millisecond)
+
+	content, readErr := os.ReadFile(path)
+	assert.NoError(t, readErr, "mount should survive broken pipe")
+	assert.Equal(t, secrets, content)
+}
```

hmnd:
@mikesellitto Thanks! I removed the `time.Sleep(100 * time.Millescond)` before the last read, as ReadFile block until MountSecrets calls OpenFile anyway. I also set maxReads=11 since we're reading it a deterministic number of times.

seslattery (approved):
Looks good to me!

emily-curry (approved):
Thanks for the contribution! This looks good to me as well.

PR #507, opened 2026-01-06
Fix CLI completions install on silicon macs

amoses12:
Release Note: Fix CLI completions installation on silicon Macs.

Closes ENG-8130

### DIFF
--- a/pkg/cmd/completion.go
+++ b/pkg/cmd/completion.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/DopplerHQ/cli/pkg/utils"
@@ -73,6 +74,8 @@ var completionInstallCmd = &cobra.Command{
 		var buf bytes.Buffer
 		var path string
 		var name string
+		// Track if using fallback path (requires manual fpath configuration)
+		var usingFallbackPath bool
 
 		if utils.IsWindows() {
 			utils.HandleError(errors.New("Completion files are not supported on Windows. You can use completion files with Windows Subsystem for Linux (WSL)."))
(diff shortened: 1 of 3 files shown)


pkg/cmd/completion.go
```diff
@@ -94,15 +96,32 @@ var completionInstallCmd = &cobra.Command{
...
+				if utils.Exists(omzDir) {
+					path = fmt.Sprintf("%s/.oh-my-zsh/completions", homeDir)
+					usingOhMyZsh = true
+				} else {
+					// Fall back to ~/.zsh/completions for non-oh-my-zsh users
+					path = fmt.Sprintf("%s/.zsh/completions", homeDir)
+				}
+			} else {
+				path = "/usr/local/share/zsh/site-functions"
+			}
```
emily-curry: nit: I'd really prefer for this to be OS-agnostic, and not have any special handling for OMZ (since it really doesn't matter). It's not guaranteed for `/usr/local/share/` to be writable on other OSes either so we need to gracefully handle when it's not.

pkg/cmd/completion.go
```diff
@@ -141,6 +175,54 @@ func getShell(args []string) string {
...
+
+	// Add newline if file doesn't end with one
+	if len(existingContent) > 0 && !strings.HasSuffix(existingContent, "\n") {
+		if _, err := f.WriteString("\n"); err != nil {
+			return err
+		}
+	}
+
+	// Write the fpath configuration block
+	configBlock := fmt.Sprintf("\n# Doppler CLI completions\n%s\nautoload -Uz compinit && compinit\n", fpathLine)
```
emily-curry: blocking: I don't trust that this will be reliable across all possible environments. Best case, we slow down every user's shell startup time by calling compinit unnecessarily, in others cases we might actually break things by doing this.

I'd much rather just print the line that needs to be added to the user's file in the case that it needs to be added manually, and expect the user to add it to the correct place.

pkg/cmd/completion.go
```diff
@@ -94,15 +96,32 @@ var completionInstallCmd = &cobra.Command{
...
+					utils.HandleError(err, "Unable to determine home directory")
+				}
+				// Check if oh-my-zsh is installed - it auto-adds ~/.oh-my-zsh/completions to fpath
+				omzDir := fmt.Sprintf("%s/.oh-my-zsh", homeDir)
+				if utils.Exists(omzDir) {
+					path = fmt.Sprintf("%s/.oh-my-zsh/completions", homeDir)
+					usingOhMyZsh = true
+				} else {
+					// Fall back to ~/.zsh/completions for non-oh-my-zsh users
+					path = fmt.Sprintf("%s/.zsh/completions", homeDir)
```
emily-curry: nit: I would not expect another program to write to this directory unless it were very well know and standardized, which is not the case here. We should be writing to our own directory.

emily-curry (changes requested):

pkg/cmd/completion.go
amoses12: @emily-curry just to clarify, we do write it to a file named "_doppler", so everything gets written to "~/.zsh/completions/_doppler". Do we want to write this to a new directory entirely or does the filename suffice?

tests/e2e/completion-install.sh
emily-curry, line 1: nit: This file isn't added to `tests/e2e.sh`, so it won't be run in CI.

emily-curry (dismissed):

emily-curry (approved):

mikesellitto (approved):
tested locally

PR #512, opened 2026-02-02
Add search for gpgv_binary common install paths if not found

amoses12:
This PR adds a search for the gpgv_binary install in common paths on linux systems before hitting the fallback logic. This is especially useful when installing in containers such as docker

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -395,7 +395,7 @@ gpgv_binary="$(command -v gpgv || true)";
 if [ -x "$gpgv_binary" ]; then
   log_debug "Using $gpgv_binary for signature verification"
 else
-  log "ERROR: Unable to find gpg binary for signature verification"
+  log "ERROR: Unable to find gpgv binary for signature verification"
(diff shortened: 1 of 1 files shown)


emily-curry (dismissed):

another-yijiao (dismissed):
could not reproduce the `command -v gpgv` failure scenario in testing but the changes make sense to me.

scripts/install.sh
```diff
@@ -392,6 +392,15 @@ fi
 log_debug "Detected format '$format'"
 
 gpgv_binary="$(command -v gpgv || true)";
+# Fallback to common paths if command -v fails (e.g., in Docker builds where PATH isn't updated)
+if [ ! -x "$gpgv_binary" ]; then
+  for path in /usr/bin/gpgv /bin/gpgv /usr/local/bin/gpgv; do
```
Piccirello: Should this really be the responsibility of `install.sh`? `/usr/bin`, `/bin`, and/or `/usr/local/bin` should be added to the system's `PATH`, which `command` will then use.

PR #514, opened 2026-02-11
Update Go version to 1.25

amoses12:


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/DopplerHQ/cli
 
 // GO_VERSION_DEF
-go 1.24.0
+go 1.25.1
```
emily-curry: check/question: Why 1.25.1 specifically?

amoses12: Good catch. I had originally read that 1.25.1 was the latest patched version for this issue, but after a little bit more digging, it should be 1.25.7. Alternatively, 1.24.13 also contains a fix for this issue if we'd like to stay on the 1.24 minor version.

emily-curry (approved):

mikesellitto (approved):

PR #517, opened 2026-03-02
Add curl timeout and retry flags to prevent install script hangs

enescakir:
The curl command in curl_download() had no connect-timeout or max-time set, causing it to hang indefinitely on slow/stalled connections. The --retry 5 flag was ineffective because curl never timed out on its own to trigger a retry.

The `dopplerhq/cli-action` action has 30 seconds timeout, so it occasionally fails without retry.

Added:
- --connect-timeout 10: fail fast if server is unreachable
- --max-time 60: cap total transfer time per attempt
- --retry-all-errors: retry on all errors, not just transient ones

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -146,7 +146,7 @@ curl_download() {
 
   # allow curl to fail w/o exiting
   set +e
-  headers=$(curl --tlsv1.2 --proto "=https" -w "%{http_code}" --silent --retry 5 -o "$output_file" -LN -D - "$url" 2>&1)
+  headers=$(curl --tlsv1.2 --proto "=https" -w "%{http_code}" --silent --connect-timeout 10 --max-time 60 --retry 5 --retry-all-errors -o "$output_file" -LN -D - "$url" 2>&1)
   exit_code=$?
   set -e
 


wotschofsky:
**+1** This would be very helpful for us. We regularly run into ETIMEDOUT errors, making CI fail.

amoses12:
Hey @enescakir, I've reviewed and tested this and this all looks good. Could you please remove claude as a co-author for the commit? Once that's done, I'll approve this and get it merged. Thanks!

enescakir:
> Hey @enescakir, I've reviewed and tested this and this all looks good. Could you please remove claude as a co-author for the commit? Once that's done, I'll approve this and get it merged. Thanks!

Thanks @amoses12, I'll remove it.

I have one more question. My main motivation for changing it was to improve the robustness of the `dopplerhq/cli-action`.

Since the action has a 30 seconds timeout, using `--max-time 60` might be too high because it prevents retries if the download stalls. The action itself times out at 30 seconds before reaching the 60 seconds max time.

However, using a much shorter `--max-time` than 60 seconds also feels a bit too aggressive. Another option could be increasing the action’s timeout or making it configurable. What do you think?

amoses12:
> > Hey @enescakir, I've reviewed and tested this and this all looks good. Could you please remove claude as a co-author for the commit? Once that's done, I'll approve this and get it merged. Thanks!
> 
> Thanks @amoses12, I'll remove it.
> 
> I have one more question. My main motivation for changing it was to improve the robustness of the `dopplerhq/cli-action`.
> 
> Since the action has a 30 seconds timeout, using `--max-time 60` might be too high because it prevents retries if the download stalls. The action itself times out at 30 seconds before reaching the 60 seconds max time.
> 
> However, using a much shorter `--max-time` than 60 seconds also feels a bit too aggressive. Another option could be increasing the action’s timeout or making it configurable. What do you think?

Hey @enescakir sorry for the late reply. I would agree that a 60 second max time is appropriate for the CLI, so I don't think it's necessary to make it configurable. Increasing the action's timeout to better support this value seems reasonable as well, it just needs to happen in the action itself. I'll go ahead and approve this PR.

amoses12 (approved):

emily-curry (approved):

PR #540, opened 2026-07-24
Migrate goreleaser off deprecated brews and dockers

watsonian:
`goreleaser check` fails on the v2.16 `brews` deprecation, and `dockers`/ `docker_manifests` are being phased out in favor of `dockers_v2`.

- Replace `brews` with `homebrew_casks` (binaries + completions, plus a post-install hook to clear the macOS quarantine attribute since the binaries are GPG-signed but not Apple-notarized).
- Replace `dockers` with `dockers_v2`, preserving the plain single-arch linux/amd64 image and its labels (sbom/provenance left off to match the previous output).
- Update docker/alpine to COPY from $TARGETPLATFORM, as dockers_v2 stages artifacts per-platform.
- Set up Docker Buildx in the release workflow, which dockers_v2 requires.
- Bump goreleaser-action from v6 to v7 in the build and release-test workflows.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -105,28 +105,29 @@ changelog:
     - Merge pull request
     - Merge branch
 
-dockers:
+dockers_v2:
   - dockerfile: docker/alpine
-    goos: linux
-    goarch: amd64
     ids:
       - doppler
-    image_templates:
-      - "dopplerhq/cli:{{ .Version }}"             # Ex: 1.4.2
-      - "dopplerhq/cli:{{ .Major }}.{{ .Minor }}"  # Ex: 1.4
-      - "dopplerhq/cli:{{ .Major }}"               # Ex: 1
-      - "dopplerhq/cli:latest"
-      - "gcr.io/dopplerhq/cli:{{ .Version }}"             # Ex: 1.4.2
-      - "gcr.io/dopplerhq/cli:{{ .Major }}.{{ .Minor }}"  # Ex: 1.4
-      - "gcr.io/dopplerhq/cli:{{ .Major }}"               # Ex: 1
-      - "gcr.io/dopplerhq/cli:latest"
-    build_flag_templates:
-      - "--label=org.label-schema.schema-version=1.0"
-      - "--label=org.label-schema.version={{.Version}}"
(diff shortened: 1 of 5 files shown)


.goreleaser.yml
```diff
@@ -105,45 +105,55 @@ changelog:
...
+    tags:
+      - "{{ .Version }}"        # Ex: 1.4.2
+      - "{{ .Major }}.{{ .Minor }}"  # Ex: 1.4
+      - "{{ .Major }}"          # Ex: 1
+      - "latest"
+    labels:
+      org.label-schema.schema-version: "1.0"
+      org.label-schema.version: "{{ .Version }}"
+      org.label-schema.name: "{{ .ProjectName }}"
+    # keep publishing a plain single-arch image, matching the pre-v2 behavior.
```
emily-curry: nit: Can you please remove any comments that an LLM included to describe this change? This line in particular only makes sense in the context of this PR, it would not make sense reading the file once it landed in master (it gives the impression that there is some important reason we're not publishing other images, when that's not true and it's just an artifact of this specific change).

emily-curry (dismissed):

.goreleaser.yml
```diff
@@ -105,45 +105,55 @@ changelog:
...
+homebrew_casks:
   - name: doppler
     repository:
       owner: DopplerHQ
       name: homebrew-doppler
     commit_author:
       name: "Doppler Bot"
       email: bot@doppler.com
-    directory: Formula
     homepage: "https://doppler.com"
```
emily-curry, line 138: check: Since we set `homepage`, I think we'll need to set `url` as well. See https://goreleaser.com/resources/deprecations/#brews

watsonian:
@emily-curry Okay, should be good. Once we merge this, can we trigger a rebuild to generate the cask or do we need to trigger a new release?

emily-curry:
> @emily-curry Okay, should be good. Once we merge this, can we trigger a rebuild to generate the cask or do we need to trigger a new release?

We'll need to cut an entirely new release, we can't cut releases for tags we've already released.

emily-curry (approved):

PR #548, opened 2026-08-14
Disable analytics by default

### LINKED ISSUE #539: [BUG] `doppler completion` makes a network request on every run, slowing shell startup
**Describe the bug**

The docs and `doppler completion --help` recommend sourcing completions live from your shell rc:

```zsh
source <(doppler completion)   # zsh; bash is equivalent
```

The completion script is static text, but running `doppler completion` performs a network request every time : the anonymous analytics POST (and, when enabled, the update check).

Because it's sourced on every interactive shell start, **shell startup blocks on that request**. On a high-latency or flaky connection the delay is very noticeable, on every new shell/tab/split.

`--no-check-version` does not fix it: the analytics POST still fires. There appears to be no flag that makes `completion` fully offline.

**To Reproduce**

```
$ doppler --no-check-version completion --debug 2>&1 >/dev/null | grep -i http
Debug: Sending anonymous analytics payload: '{"command":"doppler.completion"}'
Debug: Performing HTTP POST to https://cli.doppler.com/v1/analytics
```

(Same analytics POST documented for `secrets get` in #511.) To see startup impact:

```
strace -f -e trace=connect zsh -i -c exit 2>&1 | grep AF_INET
```

**Expected behavior**

`doppler completion` should never make a network call: its output is static per CLI version.

**Screenshots**

N/A

 - OS: Arch Linux
 - Version: zsh 5.9

**CLI Version:**
Version 3.76.0

**Additional context**

N/A

emily-curry:
We're leaving all the plumbing in place, so a future change can better handle opt-ins.

Closes #539 
Closes ENG-9607

### DIFF
--- a/pkg/configuration/flags.go
+++ b/pkg/configuration/flags.go
@@ -58,7 +58,7 @@ func SetFlag(flag string, enable bool) {
 func GetFlagDefault(flag string) bool {
 	switch flag {
 	case models.FlagAnalytics:
-		return true
+		return false
 	case models.FlagEnvWarning:
 		return true
 	case models.FlagUpdateCheck:
--- a/pkg/models/config.go
+++ b/pkg/models/config.go
@@ -45,8 +45,7 @@ type VersionCheck struct {
 }
 
 type AnalyticsOptions struct {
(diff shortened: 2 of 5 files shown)


mikesellitto (approved):
👍🏻

PR #552, opened 2026-08-26
Pin GitHub Actions to commit SHAs

mikesellitto:
Closes ENG-9517

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,12 +7,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Set up Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -7,12 +7,12 @@ jobs:
...
     steps:
     - name: Set up Go
-      uses: actions/setup-go@v3
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
       with:
         go-version: '1.25' # GO_VERSION_DEF
         check-latest: true
     - name: Checkout
-      uses: actions/checkout@v1
+      uses: actions/checkout@50fbc622fc4ef5163becd7fab6573eac35f8462e # v1.2.0
```
emily-curry, line 15: check: Why are we pinning these at such an old version?

mikesellitto: Since this was already `@v1` I figured it safest to pin to what was already running since I haven't looked into what would be required to upgrade to a newer major version

emily-curry (approved):

PR #555, opened 2026-09-09
Add tags commands

mikesellitto:
This PR adds a `doppler tags` command group which will call our v3 tags endpoints once supported.

Note: Not to be merged until feature flag removal

Specifically, this PR:
- Adds `doppler tags` which lists tags
  - Adds `get`, `create`, `update`, and `delete` subcommands
    - `create` has optional flags `--color` and `--slug`
    - `update` requires at least one of `--name` or `--color`
    - `delete` prompts for confirmation unless `-y` is passed
- Adds a `PatchRequest` helper since the update endpoint is a `PATCH` and we only had `PUT`

Closes ENG-9686

### DIFF
--- a/pkg/cmd/tags.go
+++ b/pkg/cmd/tags.go
@@ -1,0 +1,208 @@
+/*
+Copyright © 2026 Doppler <support@doppler.com>
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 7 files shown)


pkg/cmd/tags.go
```diff
@@ -0,0 +1,208 @@
...
+	jsonFlag := utils.OutputJSON
+	color := cmd.Flag("color").Value.String()
+	slug := cmd.Flag("slug").Value.String()
+	localConfig := configuration.LocalConfig(cmd)
+
+	name := args[0]
+
+	utils.RequireValue("token", localConfig.Token.Value)
+	utils.RequireValue("name", name)
+
```
amoses12, line 123: nit: we should probably check to see if the color is actually a valid color here in the handler as opposed to sending it off to the server and having it error there.

pkg/cmd/tags.go
```diff
@@ -0,0 +1,208 @@
...
+	jsonFlag := utils.OutputJSON
+	name := cmd.Flag("name").Value.String()
+	color := cmd.Flag("color").Value.String()
+	localConfig := configuration.LocalConfig(cmd)
+
+	slug := args[0]
+
+	utils.RequireValue("token", localConfig.Token.Value)
+	utils.RequireValue("slug", slug)
+
```
amoses12, line 144: nit: same here, let's check for a valid color before making the round trip to the server.

amoses12 (changes requested):
Feel free to push back on these, but it's a small optimization and we use this pattern elsewhere in the CLI. Otherwise this looks good.

